PR #16, opened 2026-05-19
[ASK] Create lambda API

jachan:
## Link to ticket
https://github.com/newjersey/tax-relief-status-checker/issues/25

## LLM Disclaimer
Claude Code was used to generate the code for the API logic. Tests were written manually.

## What was done?
- Add Lambda to query DORES database
- Add script to zip the lambda up with minimal dependencies (remove devDependencies before zipping)

## Accessibility
N/A, backend changes

## Steps to test
- `npm run test`

N/A

## Notes
- Requires config file, which has been not checked into this public repo.
- Would love scrutiny/suggestions around what to leave public vs private here
- Tested this by deploying to AWS (using `npm run package` and then uploading the zip) and testing the following:
  - malformatted requests
  - properly formatted request with no corresponding filer (returns 200 w empty filer)
  - properly formatted request with a corresponding filer (returns 200 w filer info)


vite.config.ts
```diff
@@ -10,6 +10,7 @@ export default defineConfig({
     environment: "jsdom",
     globals: true,
     setupFiles: "./vitest.setup.ts",
+    exclude: ["lambda/**", "node_modules/**"],
```
jviall, line 13: [sand] Why exclude `node_modules`? My understanding is Vite is pretty smart about this.

lambda/index.test.js
```diff
@@ -0,0 +1,305 @@
...
+  const mockSend = jest.fn().mockResolvedValue({
+    SecretString: JSON.stringify({
+      ORACLE_DB_USER: "testuser",
+      ORACLE_DB_PASSWORD: "testpass",
+    }),
+  });
+
+  return {
+    SecretsManagerClient: jest.fn(() => ({ send: mockSend })),
+    GetSecretValueCommand: jest.fn(),
```
jviall: [sand] You'll have a much better time testing using [aws-sdk-client-mock](https://github.com/m-radzikowski/aws-sdk-client-mock) to mock out the SDK commands.

lambda/index.js
```diff
@@ -0,0 +1,175 @@
+const { SecretsManagerClient, GetSecretValueCommand } = require("@aws-sdk/client-secrets-manager");
+const oracledb = require("oracledb");
```
jviall: [sand] Is there a plan to move from Common JS syntax to ESModule + TypeScript.

I imagine you wanted to just get a proof of concept up and running, but fwiw a `cdk init` is pretty easy and straightforward to setup a CDK stack for a singular Lambda straight from the get go (Claude can definitely handle bootstrapping it for you too), and takes care of lots of the typescript and bundling side of things. Certainly will make your packaging/uploading steps much simpler and future-proof!

lambda/index.js
```diff
@@ -0,0 +1,175 @@
...
+const MAX_TRANSACTION_SLOTS = 6;
+
+/** Maps TRANS_X_TAX_CDE values to response category names */
+const TAX_CODE_CATEGORY_MAP = {
+  13: "anchor",
+  49: "ptr",
+  41: "stayNj",
+};
+
+const validateInput = (event) => {
```
jviall: [dust] Since none of these helpers are exported, they can't be individually unit-tested in an isolated way. Would probably be a big improvement to the unit tests down the line.

lambda/index.js
jachan: Thanks! Creating a CDK stack to define infra was going to be my next PR. Let me see what I can do with Claude to migrate to ESModule + TS though, I feel like that's definitely a better place for this PR to be over just using default Common JS Syntax (which I just pulled from AWS docs)

lambda/index.test.js
jachan: Appreciate the pointer!

vite.config.ts
jachan: Without this line to exclude `node_modules/**`, `vitest` tries to run test files within `node_modules`.

I tried alternatively updating `vitest.setup.ts`, but it fields unwieldy to update both a local lambda `vitest.setup` AND a root level `vitest.setup` (we'd need two since the root level `vite.config.ts` ignores the lambda subdirectory).

jviall: Fwiw, the separate configs is a pretty common pattern, and with the introduction of this lambda you're veering into monorepo-ish land. The lambda you're standing up is a very distinct stack from the web app, and it makes sense that they would have distinct build and test setups. If there is redundant configurations between multiple projects, it's simple enough to have a shared root config, and project-specific configs that make necessary customizations.

For us on CCM (where we have many lambdas and a complex CDK stack), we only need vitest configs and tsconfigs because the CDK handles bundling of the lambdas, but we have these configs for each "project" of the repo. In your case, this would likely end up being one set of configs for your webapp, and another for your lambda. Since the repo is fairly simple though, you may not need to fully drink the monorepo cool-aid and get away with a somewhat simpler config setup. I would still advocate for a separate vitest config though, and then let the webapp's `vite` config completely ignore the lambda folder.

Happy to pair on this if you would appreciate help!

lambda/src/index.test.ts
```diff
@@ -0,0 +1,175 @@
+import { describe, it, expect, beforeEach, vi } from "vitest";
+import { mockClient } from "aws-sdk-client-mock";
+import { SecretsManagerClient, GetSecretValueCommand } from "@aws-sdk/client-secrets-manager";
+
+import { handler } from "./index.ts";
+
+const secretsMock = mockClient(SecretsManagerClient);
```
jviall, line 7: [praise] much cleaner throughout 🥇

vite.config.ts
jviall: > vitest tries to run test files within node_modules

Also I think what's happening here is that Vite/Vitest has sane defaults for the `exclude` config, but when you override it with something those defaults (which definitely exclude `node_modules`) get wiped. Having separate configs that instead are explicit about what they _**include**_, would also solve this issue.

jachan: Heard. the vision for this project was to be a monorepo with webapp, lambda, and infra all defined in the same place. Understand the desire for separate configs per subdirectory. In the spirit of a "sand", I'm gonna handle this in the followup PR where I migrate to CDK defined infra.

package.json
```diff
@@ -8,7 +8,7 @@
   "scripts": {
     "dev": "vite",
     "build": "tsc -b && vite build",
-    "test": "vitest",
+    "test": "vitest run && cd lambda && npm i && npm run test",
```
jviall, line 11: [pebble] I think Claude is leading you astray here with this nested package.json. You're only halfway committing to a monorepo setup by having this extra pakcage.json without formally creating an npm workspace--but it's not clear to me that it's strictly necessary to do either, you could probably get away with one root package.json that just had additional commands to specifically manage the lambda.

It's also weird to me for an `npm` script to run `npm install` inside of itself.

jviall (commented):
I think there's still some things to iron out on the project setup stuff. However, it might be fair to cleanly separate these efforts out, and just try to land your source code in whatever simplest form actually works.

Consider this a soft approval!

vite.config.ts
jviall: > the vision for this project was to be a monorepo

in that case, the CCM project might serve as a helpful example!

package.json
jachan: James! That we ME leading ME astray, not Claude! But thankfully you have set me back on the right path and ordered my steps. I shall commit to full monorepo and all will be well.

lambda/src/index.test.ts
jachan: AI will never take MY job

package.json
jviall: homer-dissapear.gif

lambda/src/index.test.ts
```diff
@@ -0,0 +1,175 @@
...
+    expect(result.statusCode).toBe(400);
+    expect(JSON.parse(result.body).error).toBe("SSN must be 9 digits");
+  });
+
+  it("returns 400 for invalid ZIP (not 5 digits)", async () => {
+    const result = await handler({ ssn: "123456789", zip: "123" });
+
+    expect(result.statusCode).toBe(400);
+    expect(JSON.parse(result.body).error).toBe("ZIP must be 5 digits");
+  });
```
aloverso, line 82: [dust] If you wanted to, a pattern to consider making these easier to test would be to extract your validation out into a pure domain function.

It's kind of tedious to write a lot of tests for all the different cases of invalid ZIPs and whatnot - so your approach here of having just one test that says "yea the validation is occurring in the handler" makes sense.

For additional confidence, if the validator was extracted, you could also have a separate test that calls that pure function without any need for all the API setup and all that. Then you could have a bunch of tests. Like zip too long (not just too short) or missing altogether. As well as a "returns true when zip is 5 digits". It would just be a super lightweight extra confidence.

Optional, hence the dust, but I think it's a good pattern to consider not just here but also in general when these kinds of nitty-gritty logic are getting tested at a high level (like in this case, small validations getting tested by a full integration-level test)

lambda/src/index.test.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  it("accepts SSN with hyphens and strips them", async () => {
```
aloverso, line 84: [pebble] I think you're missing an expectation here for confirming the "and strips them" part of this

lambda/src/index.test.ts
```diff
@@ -0,0 +1,175 @@
...
+      const body = JSON.parse(result.body);
+
+      expect(body["records"]).toHaveLength(1);
+      expect(body["records"][0].return_year).toBe("2024");
+      expect(body["records"][0].application_date).toBe("10/31/2025 00:00:00");
+    });
+  });
+
+  describe("when filer has multiple matching rows in DB", () => {
+    it("eturns 200 and constructs the response", async () => {
```
aloverso: [pebble] typo

Issue #25, opened 2026-05-20
Build API & Infrastructure


emtwelp:

jachan:
https://github.com/newjersey/tax-relief-status-checker/pull/16/ merged - will work on infrastructure next

jachan:
Infrastructure implemented in https://github.com/newjersey/tax-relief-status-checker/pull/57. Ticket is complete.

PR #54, opened 2026-05-26
[ASK] Refactor to a monorepo

jachan:
## Link to ticket
https://github.com/newjersey/tax-relief-status-checker/issues/25

## What was done?
- Created monorepo structure to enable infra to be a separate folder in this repo

## LLM Disclaimer
No LLMs were used in the making of this PR

## Accessibility
N/A, involves no frontend changes

## Steps to test
- ran all package.json scripts and made sure that repo was still functional

N/A

## Notes
First time creating a monorepo, so all feedback is welcome!


package.json
```diff
@@ -1,45 +1,35 @@
...
-  }
+  "author": {
+    "name": "New Jersey State Office of Innovation",
+    "url": "https://github.com/newjersey"
+  },
+  "license": "MIT",
+  "workspaces": [
+    "webapp",
+    "lambda"
+  ]
```
seidior, line 34: [dust] It's not clear what kind of workspaces you're setting up. Are we using `npm` [workspaces](https://docs.npmjs.com/cli/v11/using-npm/workspaces) (which, by convention, use `packages/` as a subfolder), `yarn` [workspaces](https://yarnpkg.com/features/workspaces), or `pnpm` [workspaces](https://pnpm.io/workspaces)?

webapp/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "homepage": "https://github.com/newjersey/tax-relief-status-checker#readme",
+  "bugs": {
+    "url": "https://github.com/newjersey/tax-relief-status-checker/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/newjersey/tax-relief-status-checker.git"
+  },
+  "license": "MIT",
+  "author": "New Jersey State Office of Innovation",
```
seidior, line 14: [dust] You may want to use the latest version of `package.json` [in the template repo](https://github.com/newjersey/innovation-repo-template/blob/main/package.json), which uses NJIA as the author instead of Office of Innovation

webapp/package.json
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "dependencies": {
+    "@newjersey/njwds": "^2.9.1",
+    "@trussworks/react-uswds": "^11.0.0",
+    "@vitejs/plugin-react": "^6.0.1",
+    "jsdom": "^29.1.1",
+    "react": "^19.2.4",
+    "react-hook-form": "^7.72.1",
+    "react-router-dom": "^7.15.1",
+    "typescript": "^6.0.3"
```
seidior: [sand] You may want to pin specific dependency versions rather than version ranges. This will help prevent issues with newer, potentially-compromised package versions.

seidior (commented):
Love a good code reorganization, but not sure if I should be Pickaroo'd on this repo! Otherwise, I like the work here, and I just have a couple of questions.

webapp/.prettierrc
sannidhishukla, line 1: [dust] what's the reason for moving the prettier config files to the `webapp` directory instead of leaving them in the root directory? will prettier not be used for the `lambdas` directory? if not, why is that (out of curiosity)?

sannidhishukla (approved):
looks good! added one question!

package.json
jachan: `npm` workspaces, using `packages/` feels like it'd be confusing here so i avoided it.

webapp/.prettierrc
jachan: Great point! The same linting rules should apply to both directories, let me try and fix that.

jachan:
@seidior the repo was misconfigured to use "innovation-engineering" GH team rather than "innovation-resx-engineering". Fixed!

Issue #26, opened 2026-05-20
Define Infrastructure using Infrastructure As Code IAC (Amazon CDK)


emtwelp:

jachan:
Hi Spencer! That ticket seems to have done the trick, Dev is now connected :)

PR #57, opened 2026-05-28
Define IAC

jachan:
## Link to ticket

#26 

## LLM Disclaimer

- Claude was used to generate the first draft of the IAC code

## What was done?

- Defined Lambda, IAM policies, and configuration for dev and prod using Amazon CDK

## Accessibility
N.A

## Steps to test

- This was tested by synthesizing the template in the AWS account and making sure that the resulting API functioned as expected. 

N/A

## Notes
`npm audit` reporting medium and high vulnerabilities that are specific to cdk library, unsure how to fix, as it breaks build when I upgrade dependency for CDK


infra/lib/infra-stack.ts
```diff
@@ -0,0 +1,116 @@
...
+import * as path from "path";
+import { fileURLToPath } from "url";
+
+const DIRNAME = path.dirname(fileURLToPath(import.meta.url));
+
+/** Duration in seconds before the Lambda function times out. */
+const LAMBDA_TIMEOUT_SECONDS = 5;
+
+/** Memory allocated to the Lambda function in megabytes. */
+const LAMBDA_MEMORY_MB = 512;
```
ezhangy: [dust] IMO the const names are self-explanatory enough that the inline comments are unnecessary

infra/lib/infra-stack.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    this.lambdaFunction.role?.attachInlinePolicy(customPolicy);
+  }
+
+  /**
+   * Builds VPC configuration for the Lambda function. When a VPC is provided, places the Lambda in
+   * private subnets and creates a Secrets Manager VPC endpoint.
+   */
+  private buildVpcConfig(
+    vpc: ec2.IVpc,
```
ezhangy, line 63: [dust] the function body seems to presume `vpc` is a required parameter

if that's the intended behavior, I found the docs comment to be a bit misleading. Reading "When a VPC is provided..." made me think the parameter was going to be optional

infra/lib/stage-config.ts
```diff
@@ -0,0 +1,47 @@
+/** Defines the deployment stages and their associated AWS account configuration. */
```
ezhangy, line 1: [dust] also think this inline comment is unnecessary

infra/lib/stage-config.ts
```diff
@@ -0,0 +1,47 @@
...
+  readonly vpcId: string;
+
+  /** The key of the credentials secret in AWS Secrets Manager */
+  readonly credentialsName: string;
+
+  /** The arn for the credentials secret in AWS Secrets Manager */
+  readonly secretArn: string;
+}
+
+/** Stage configurations keyed by stage name. */
```
ezhangy, line 27: [dust] same with this inline comment

infra/tsconfig.json
```diff
@@ -0,0 +1,26 @@
+{
```
ezhangy, line 1: [dust] FYI according to the NJ AI assistant, the `noImplicitAny, strictNullChecks, noImplicitThis, alwaysStrict` options are redundant since they are already enabled by `"strict": true`

but I don't have a particular opinion on whether they should be removed

infra/lib/infra-stack.ts
```diff
@@ -0,0 +1,116 @@
...
+export interface InfraStackProps extends StackProps {
+  readonly stageName: string;
+  readonly vpcId: string;
+  readonly connectString: string;
+  readonly credentialsName: string;
+  readonly secretArn: string;
+}
+
+/**
+ * CDK stack defining the Property Tax Relief Status API Lambda function. Supports optional VPC
```
ezhangy: [dust] similar comment about how the comment says that the VPC is optional, but it seems required based on the code

infra/lib/infra-stack.ts
```diff
@@ -0,0 +1,116 @@
...
+        DB_CREDS_SECRET_NAME: props.credentialsName,
+      },
+      ...vpcConfig,
+    });
+
+    const customPolicy = new iam.Policy(this, "LambdaCustomPolicy", {
+      statements: [
+        new iam.PolicyStatement({
+          actions: ["secretsmanager:GetSecretValue"],
+          resources: [props.secretArn],
```
ezhangy: [dust] out of curiosity, what does secret does the `secretArn` point to? is it useful to document that in the property name (e.g. `dbPasswordSecretArn`)? 

right now the code doesn't really give a hint to what secret the policy is permits the Lambda to access

infra/lib/infra-stack.ts
```diff
@@ -0,0 +1,116 @@
...
+  private buildVpcConfig(
+    vpc: ec2.IVpc,
+  ): Pick<lambda.FunctionProps, "vpc" | "vpcSubnets" | "securityGroups"> {
+    const securityGroup = new ec2.SecurityGroup(this, "LambdaSecurityGroup", {
+      vpc,
+      description: "Security group for the Property Tax Relief Status API Lambda",
+      allowAllOutbound: true,
+    });
+
+    const endpointSecurityGroup = new ec2.SecurityGroup(
```
ezhangy: [dust] IMO this is more personal preference, but the variable names for the security groups are a bit generic

something like `lambdaSecurityGroup` and `secretsManagerEndpointSecurityGroup` would be clearer

ezhangy (approved):
looks great!! left several dusts, mostly small nitpicky stuff!

infra/tsconfig.json
jachan: Removed! No need for redundant options IMO

Issue #58, opened 2026-05-29
Build lookup page


emtwelp:
**Context:** Residents land on this page when they want to check the status of their property tax relief application. It's the entry point to the tool.

**Designs:**
- [State 1: Before submit](https://www.figma.com/design/fc59D8BV14x7EhW79ued5f/PAS-1-status-checker?node-id=312-4508&t=fMnKrr8znSmt5hvX-4)
- State 2: After submit if no application found
- State 3: System error

**Acceptance Criteria:**
- If no application is found, error message appears on the lookup page.
- Error message includes a link that scrolls to the FAQ section via anchor link.
- FAQ is implemented as an accordion: first section is expanded by default, remaining sections start collapsed.

jachan:
Completed in https://github.com/newjersey/tax-relief-status-checker/pull/74

PR #68, opened 2026-06-04
58 - Update Landing Page

jachan:
## Link to ticket
#58 

## LLM Disclaimer
An LLM was used to generate the table in the FAQ section titled "When can I expect to receive payments?"

## What was done?
- Made content match [Figma Mocks](https://www.figma.com/design/fc59D8BV14x7EhW79ued5f/PAS-1-status-checker?node-id=392-2818&t=N4la52eGJPDaCNGn-0)
- Added `react-hook-form` validation to form fields

### OUT OF SCOPE: 

- Adding Error messages tied to API call, and adding API call. There was enough complexity with the API logic that I decided to break that into a second ticket.
- Adding Feedback Widget will be handled in a separate ticket

## Accessibility

## Steps to test

- `npm run dev` and inspect to make sure it matches the mocks from the [Figma](https://www.figma.com/design/fc59D8BV14x7EhW79ued5f/PAS-1-status-checker?node-id=392-2818&t=N4la52eGJPDaCNGn-0)

- Before
image

- After
image

- Figma Mocks
image


webapp/src/LandingPage.tsx
```diff
@@ -1,131 +1,123 @@
...
+      zipCode: "",
+    },
+    shouldFocusError: false,
+  });
+
+  const onSubmit: SubmitHandler<UserData> = (data) => {
+    let key: keyof UserData;
+    for (key in data) {
+      const value = data[key] ?? "";
+      window.sessionStorage.setItem(key, value);
```
jachan, line 28: This is placeholder logic until the API ticket gets implemented

webapp/src/components/LandingPageFaq.tsx
```diff
@@ -6,38 +6,168 @@ export const LandingPageFaq = (props: { headingLevel: HeadingLevel }) => {
...
+                <th>Month</th>
+                <th>Senior Freeze</th>
+                <th>ANCHOR</th>
+                <th>Stay NJ</th>
+              </tr>
+            </thead>
+            <tbody>
+              <tr>
+                <td>
+                  <strong>JUL-SEP</strong>
```
aloverso: [pebble] I checked the accessibility of this table with VoiceOver, and noticed that it reads this as "jul sep" words, not month abbreviations. I think it's generally considered good for screen readers to avoid abbreviations like this so it can be read correctly. I recommend changing to the unabbreviated month names in this table.

aloverso (commented):
Also an added [sand], I noticed while running the code that your "Return to Top" link doesn't change focus to the top if you use keyboard nav.

webapp/src/components/LandingPageFaq.tsx
```diff
@@ -6,38 +6,168 @@ export const LandingPageFaq = (props: { headingLevel: HeadingLevel }) => {
...
     },
     {
-      title:
-        "My application has been sitting in processing longer than expected, who should I reach out to?",
-      content: <p>Placeholder5</p>,
+      title: "My application is taking too long. Who do I contact?",
+      content: (
+        <p>
+          If you have any questions or if processing has taken longer than expected, call our help
+          line at 1-888-238-1233 (Mondays to Fridays 8:30 a.m. to 5:30 p.m.) or visit a Regional
```
mmazanec22: [dust] Could/should this number be an `a href tel:` to make it easier to make the call on mobile?

webapp/src/components/NjFooter.tsx
```diff
@@ -10,92 +12,22 @@ export const NjFooter = () => {
             <div className="grid-row">
               <div className="usa-footer__logo grid-row mobile-lg:grid-col-6 mobile-lg:grid-gap-2">
                 <div className="mobile-lg:grid-col-auto">
-                  <h3 className="usa-footer__logo-heading">Division of Taxation</h3>
+                  <h3 className="usa-footer__logo-heading">
```
mmazanec22: [dust] Why h3? I would think a footer header hierarchy would start with an h2

mmazanec22 (approved):
Looks good! One small potential usability change and one header level question

webapp/src/components/NjFooter.tsx
jachan: Great catch - holdover from when there was something else above it. It should be an h2.

webapp/src/components/LandingPageFaq.tsx
jachan: I don't see why not! Thanks for the suggestion.

Issue #31, opened 2026-05-20
Build real v1 website in dev environment


emtwelp:

jachan:
Resolved by https://github.com/newjersey/tax-relief-status-checker/pull/74.

v1 is accessible on Amplify Dev (see Slack for URL)

PR #69, opened 2026-06-05
[SHOW] Add lambda function url to cdk

jachan:
## Link to ticket

This commit is part of resolving https://github.com/newjersey/tax-relief-status-checker/issues/31. Future PRs are as follows:
1. Make the Lambda API endpoint accessible [YOU ARE HERE]
2. Migrate to NextJS, use API Routes and SSR to hit Lambda API endpoint

## LLM Disclaimer
No LLM was used for this code

## What was done?
We need an API endpoint for our Lambda function. This commit utilizes Lambda Function URL functionality. This was chosen over API Gateway as it's simpler to set up and tear down, and allows us to test CORS and SSR without having to set up API Gateway.

This has been deployed to dev using `npx cdk deploy -c stage=dev` so we can test SSR in a subsequent commit

Notable requirements:
- implements CORS and IAM Auth.

## Accessibility
N/A

## Steps to test
N/A

N/A

## Notes

### DIFF
--- a/infra/lib/stage-config.ts
+++ b/infra/lib/stage-config.ts
@@ -22,6 +22,9 @@ export interface StageConfig {
 
   /** The arn for the credentials secret in AWS Secrets Manager */
   readonly secretArn: string;
+
+  /** The CORS origins that are allowed to access the Lambda */
+  readonly allowedOrigins: string[];
 }
 
 /** Stage configurations keyed by stage name. */
@@ -34,6 +37,7 @@ export const STAGE_CONFIGURATIONS: Record<string, StageConfig> = {
     connectString: "TAX2_TAXU",
     credentialsName: "TAX2_CREDS_STAGING",
     secretArn: "arn:aws:secretsmanager:us-east-1:539590994798:secret:TAX2_CREDS_STAGING-eNohgJ",
+    allowedOrigins: ["*"],
   },
   prod: {
     stageName: "prod",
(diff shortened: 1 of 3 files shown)


infra/lib/stage-config.ts
```diff
@@ -34,6 +37,7 @@ export const STAGE_CONFIGURATIONS: Record<string, StageConfig> = {
     connectString: "TAX2_TAXU",
     credentialsName: "TAX2_CREDS_STAGING",
     secretArn: "arn:aws:secretsmanager:us-east-1:539590994798:secret:TAX2_CREDS_STAGING-eNohgJ",
+    allowedOrigins: ["*"],
```
jachan: Tried using the deployment Amplify URL: https://main.d1fv5gc7ba4m6a.amplifyapp.com/ but got the following error during CDK synth. Also tried using a Wildcard with the Amplify URL but that failed with the same message `https://*.amplifyapp.com`.

```
🛑 Resource handler returned message: "https://main.d1fv5gc7ba4m6a.amplifyapp.com/ isn't a valid origin. An origin must be in a valid URL format. For example: https://www.example.com, https://*, or the wildcard character (*).
```

Will specifying `amplifyapp.com` work as the origin also allow all subdomains? 🤔 [Docs](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.FunctionUrlCorsOptions.html) are pretty unclear about this. Was just going to try different things and see what works after I get SSR set up and NextJS migration done.

PR #74, opened 2026-06-08
31 Migrate to NextJS, add SSR API route, add API logic

jachan:
## Link to ticket
#31 

## LLM Disclaimer

- An LLM was used to create the initial migration from Vite to NextJS, this code was then manually reviewed and extraneous changes were removed
- An LLM was used to generate test code for the helper functions, as well as write guidance to make future unit tests less verbose

## What was done?

- Migrated to NextJS so we could take advantage of Server Side API Routes. This allows us to authenticate requests with AWS IAM by signing them server-side before they go to our lambda. 
- Add API route that serves as a proxy for requests before they go to our lambda
- Update page to add a `onSubmit` handler that actually makes an API request through the SSR API route

## Accessibility

## Steps to test
This has been deployed at https://31-migrate-to-nextjs.d1jwhothjet40j.amplifyapp.com/. Testing locally will cause all API requests to fail, as local reviewers do not have the permissions to call the API hosted in the dev account. DM John for credentials to access this site, as it is password protected.

The Amplify Deployment, however, is configured with the IAM Role necessary.

Reviewers can test with the following test data:
1) SSN: 111-11-1111 and Zip 22222 will have no records found
2) SSN: 001-24-7873 and Zip 07016 is a test record in the database that should take you to the status page.

- Before
image
image

- After
image

Error messages:

4XX error and 5XX error
image

200 with no record found
image

__

API now works and takes you to the status page)
image

- Mocks ([LINK TO FIGMA](https://www.figma.com/design/fc59D8BV14x7EhW79ued5f/PAS-1-status-checker?node-id=391-2167&p=f&t=jQHgmNGh4RjaTl93-0))
image

image


webapp/components/LandingPageFaq.tsx
jachan, line 1: File moved and mostly unchanged except for adding helper function `expandFaqAccordionItem`, a11y discussion here on this approach (hyperlink and javascript): https://www.figma.com/design/fc59D8BV14x7EhW79ued5f?node-id=392-2923#1784212479

webapp/components/HorizontalDivider.tsx
jachan, line 1: File moved and unchanged.

webapp/components/NjFooter.tsx
jachan, line 1: File moved and unchanged.

webapp/components/NjHeader.tsx
jachan, line 1: File moved and unchanged.

webapp/.env.local.example
jachan, line 1: This exists as documentation to show that an environment variable needs to be set in the Amplify console to have this application work properly.

CDK doesn't appear to support amplify configuration of Environment Variables rn: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-amplify-alpha-readme.html

sannidhishukla:
@jachan where can i find the login info to access the amplify site you linked in the PR description?

webapp/components/LandingPageFaq.tsx
sannidhishukla, line 1: [dust] IMO the newly added code comments on this page could probably be removed without causing confusion

webapp/app/status/page.tsx
sannidhishukla, line 1: [question] - right now, when i log into the page, copy the URL with search params, log out, and then paste the URL with search params into the browser, i'm still able to see the logged in view. 

i'm assuming this is behavior that was set up just to build out a short term proof of concept but i'm curious what the future expected login / logout behavior here will look like if you already have a sense of that?

webapp/app/page.tsx
```diff
@@ -0,0 +1,224 @@
...
+            </p>
+          </>,
+        );
+        return;
+      }
+
+      const maskedSsn = maskSsn(data.ssn);
+      const formattedDate = formatDate(record2025.application_date);
+      const params = new URLSearchParams({
+        ssn: maskedSsn,
```
sannidhishukla: [dust] is there a reason for using the masked SSN in the query params instead of just the last 4 digits by themselves?

sannidhishukla (approved):
looks good! left some questions!

webapp/app/page.tsx
jachan: No, thanks for suggesting that! I like how much simpler it is.

webapp/app/status/page.tsx
jachan: Yes! Right now for V1 we are grabbing minimal information from the DB. With V2 and V3 of the site, the response from the API will become a lot more complex, and with that launch, I think it'll make sense to move away from this "query params" => render the page approach that we're doing here. 

But for V1, I think it's a pretty straightforward implementation that saves us from having to store a lot of state in the webapp, which I why I chose to do this for now.

webapp/app/page.tsx
```diff
@@ -0,0 +1,224 @@
...
+              className="usa-alert usa-alert--error usa-alert--slim margin-bottom-3"
+              role="alert"
+            >
+              <div className="usa-alert__body">{apiError}</div>
+            </div>
+          )}
+
+          <div className="grid-row grid-gap margin-top-5 margin-bottom-10">
+            <div className="tablet:grid-col-6">
+              <h2>This tool is for checking your 2025 PAS-1 application status</h2>
```
mmazanec22, line 147: [dust] Does this content need to be in a header?

mmazanec22 (commented):
Sorry for the late review due to being out sick! Looks good, one question about heading structure

webapp/app/page.tsx
jachan: Andrew did an a11y review for us, he suggested that this should be an H1. [[a11y review]](url) I'm going to tackle his feedback in a separate PR.

Issue #35, opened 2026-05-20
Set up analytics


emtwelp:
Ideal analytics included in [V1 Spec](https://docs.google.com/document/d/1SgTJ1BlJl-WduqCJFrQT8GmGb_nC1Q23ohgwOggffFU/edit?tab=t.0). 

Absolutely necessary for launch:
- Funnel metrics: how many people landed, looked up, got a result (and which result — found vs. not found)
- FAQ click through rates
- Feedback Widget Standard Data

PR #104, opened 2026-06-18
[SHOW] 35 - Add analytics

jachan:
## Link to ticket
https://github.com/newjersey/tax-relief-status-checker/issues/35

## LLM Disclaimer
LLM was used to debug but not to generate code.

## What was done?
Add the following metrics:
- Track when user clicks on each FAQ question
- Track when API returns 200 found vs 200 not found vs Error

## Accessibility
N/A Metrics

## Steps to test

Install Google Tag Assistant. Run this locally. Using Chrome Dev Tools, override the response to the `/api/status` route. without an override, we can simulate a 500 error.

Here are sample responses for 200s:

200 Record FOUND
```
{"records":[{"return_year":"2025","application_date":"2026-03-19T00:00:00.000Z"}]}
```

200 Record NOT FOUND
```
{"records":[]}
```

Validate that the correct metric is emitted using Google Tag Assistant.

Then, click on the FAQ accordions and validate that the correct metrics are emitted using Google Tag Assistant. These metrics should only be emitted on accordion open, not on accordion close.

API 500
image

API 200 FOUND
image

API 200 NOT FOUND
image

FAQs

image

image

image

image

image

## Notes
Need to add end to end tests as a fast-follow.
After feedback widget is deployed, ensure that we get analytics for that as well.

Paired with @natekep


package.json
```diff
@@ -14,7 +14,8 @@
     "test": "vitest run",
     "format": "prettier . --write",
     "format:check": "prettier . --check",
-    "prepare": "husky"
+    "prepare": "husky",
+    "lockfile": "rm -rf webapp/node_modules lambda/node_modules infra/node_modules node_modules package-lock.json && npm install"
```
jachan, line 18: Added helper command to regenerate the lockfile for ALL directories in the monorepo.

webapp/app/utils/analytics.ts
```diff
@@ -0,0 +1,33 @@
...
+interface EventParams {
+  user_session_id?: string;
+  timestamp: number;
+}
+
+export const setSessionId = (): void => {
+  if (sessionStorage.getItem("sessionId") == null) {
+    sessionStorage.setItem("sessionId", uuidv4());
+  }
+};
```
mluedke2, line 18: [dust] for my understanding, does gtag not already have a usable session ID? i believe you are able to track events by Unique User/Session out of the box?

webapp/app/layout.tsx
```diff
@@ -14,6 +15,20 @@ export const metadata: Metadata = {
...
+        strategy="afterInteractive"
+      />
+      <Script strategy="afterInteractive">
+        {`
+          window.dataLayer = window.dataLayer || [];
+          function gtag(){dataLayer.push(arguments);}
+          gtag('js', new Date());
+          gtag('config', 'G-KMSDMG9NFN');
+        `}
+      </Script>
```
mluedke2, line 30: [dust] for my understanding, why are both of these scripts needed?

i think there's a [sand] here to abstract the `G-KMSDMG9NFN` to one variable with a name designating it as prod or dev and/or bypassing/mocking tracking on non-prod?

webapp/app/page.tsx
```diff
@@ -30,6 +31,9 @@ const returnToTop = () => {
 
 const LandingPage = () => {
   const router = useRouter();
+  useEffect(() => {
+    setSessionId();
+  }, []);
```
mluedke2, line 36: [dust] i know it's called Landing Page, but is there any chance the user can get the URL to a later page and would end up bypassing this session id logic?

webapp/app/page.tsx
mluedke2, line 79: [dust] noting that this `setApiError()` is used for both system issues and when there's no record found, which puts it a little out of sync with the event names (which only puts `api_error` for non-ok responses). maybe something like `setApiAlert()`?

webapp/app/utils/analytics.ts
jachan: Looked into this! Looks like you're right! 

https://developers.google.com/tag-platform/gtagjs/reference, makes this a lot easier (as we don't have to worry about initializing session ID ourselves). Thanks for the suggestion!

webapp/components/LandingPageFaq.tsx
```diff
@@ -187,6 +200,7 @@ export const LandingPageFaq = (props: LandingPageFaqProps) => {
       expanded: false,
       id: "faq_application_taking_too_long",
       headingLevel: props.headingLevel,
+      handleToggle: () => fireEventWhenFaqOpened("faq_application_taking_too_long"),
```
mluedke2, line 203: [dust] the reappearance of the id here in the handler caught my eye as maybe an opportunity for DRYing if it's not onerous. maybe one way would be something like this, since `expanded` and `headingLevel` look to be the same for all too:

```
const faqData = [
  {
    id: "faq_when_can_i_expect_my_application_status",
    title: "When can I expect my application to show up?",
    content: (
      <p>
        For online applications, it can take up to 3 weeks before a status shows up. For paper
        applications, it can take up to 12 weeks before a status shows up.
      </p>
    ),
  },
  // etc
];

const faqContent: AccordionItemProps[] = faqData.map(({ id, title, content }) => ({
  title,
  content,
  expanded: false,
  id,
  headingLevel: props.headingLevel,
  handleToggle: () => fireEventWhenFaqOpened(id),
}));
```

mluedke2:
added some comments but of course take or leave as makes the most sense! looks good to me!

webapp/app/layout.tsx
natekep: Based off my understanding the scripts are needed because of a "ReferenceError: window is not defined" error that exists within NextJS. Because of the 'after-interactive' the script runs after the page is rendered ensuring that there is no headless browser and the browser globals are able to be accessed.

jachan: This is the approach instructed by Google Analytics themselves:

Image

mluedke2: got it- thanks for the explanation+reference! (honestly it's kind of wild to me that's the cleanest/simplest they can come up with but hey if it works...)

Issue #76, opened 2026-06-10
Add Feedback Widget


emtwelp:

jachan:
This is merged into main, we need to figure out when to release to prod again, and then this change can be marked as Done.

PR #111, opened 2026-06-23
76 - Introduce Feedback Widget

jachan:
## Link to ticket

#76 

## LLM Disclaimer

- An LLM was used to help resolve an issue where the component was being rendered on the server-side, leading to a failing `npm run build`

## What was done?

- Added dependency on @newjersey/feedback-widget
- Added feedback widget to footer of all pages

## Accessibility

WAVE reports error
image

## Steps to test

- Describe how a reviewer should test the changes, including: 
  - Feedback Widget does not work locally, as `localhost` is a blocked domain, so requests to the feedback widget API fail with a CORS error.
  - Feedback Widget has been deployed to https://76-feedback-widget.d1fv5gc7ba4m6a.amplifyapp.com/, DM @jachan  for username/password to staging site.

- Before
image

- After
image

### DIFF
--- a/webapp/components/FeedbackWidget.tsx
+++ b/webapp/components/FeedbackWidget.tsx
@@ -4,8 +4,7 @@ import { useRef, type ReactElement } from "react";
 import "@newjersey/feedback-widget/feedback-widget.min.js";
 
 // NJ: Tells TypeScript that <feedback-widget> is a valid custom element.
-declare global {
-  //eslint-disable-next-line @typescript-eslint/no-namespace
+declare module "react" {
   namespace JSX {
     interface IntrinsicElements {
       "feedback-widget": React.DetailedHTMLProps<React.HTMLAttributes<HTMLElement>, HTMLElement>;
@@ -22,7 +21,7 @@ export const FeedbackWidget = (): ReactElement => {
         <feedback-widget
           ref={widgetRef}
           contact-link="https://www.nj.gov/treasury/taxation/contact.shtml?open=email"
-        />
+        ></feedback-widget>
       </section>
     </>
   );
--- a/webapp/components/FeedbackWidget.tsx
+++ b/webapp/components/FeedbackWidget.tsx
@@ -1,7 +1,7 @@
 "use client";
 
-import { useRef, type ReactElement } from "react";
-import "@newjersey/feedback-widget/feedback-widget.min.js";
+import { type ReactElement } from "react";
+import dynamic from "next/dynamic";
 
 // NJ: Tells TypeScript that <feedback-widget> is a valid custom element.
 declare module "react" {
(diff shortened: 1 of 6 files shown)


mmazanec22 (approved):
Seems straightforward! Hello feedback widget!

webapp/components/FeedbackWidget.tsx
```diff
@@ -0,0 +1,30 @@
...
+}
+
+// Allows us to render this dynamically to prevent server side rendering
+const FeedbackWidgetClientOnly = (): ReactElement => {
+  import("@newjersey/feedback-widget/feedback-widget.min.js");
+
+  return (
+    <>
+      <section aria-label="Leave Feedback">
+        <feedback-widget contact-link="https://www.nj.gov/treasury/taxation/contact.shtml?open=email"></feedback-widget>
```
ezhangy, line 22: [sand] how much traffic are you expecting to the feedback widget? 

if it's on the order of thousands of responses per month, I would recommend setting `only-save-rating-to-analytics="true"` 

(this will however require setting up Google Analytics, and making sure the custom rating events are configured properly in the case that you're using Google Tag Manager)

ezhangy (approved):
looks good! left a sand about `only-save-rating-to-analytics` attribute that's definitely worth considering before going to prod, but not blocking for this PR

webapp/components/FeedbackWidget.tsx
jachan: Our Google Analytics is set up - how do we configure this to hit our own GA4 property for Tax Status Checker? Or will this always report to a different GA4 account?

ezhangy: There are two situations, whether you're using 1) GA without GTM or 2) GTM.

**GA without GTM**
You don't have to configure the feedback widget specifically to your GA4 property. If the page is already set up to send events to your GA4 property, then the custom GA events fired by the widget will automatically there too. 

In this case, the event name would be `click_feedback_rating` and the rating data (`Yes` or `No`) would be under the dimension `object_type`. 

All you need to do to start tracking the events is add the `object_type` custom dimension to your GA4 property (see [Step 4 - add custom dimensions to GA4](https://newjersey.github.io/innovation-engineering/guides/monitoring/setting-up-google-analytics/#step-4---add-custom-dimensions-to-ga4-if-necessary) in the wiki)

**GTM**
However, configuration _is_ necessary if you've set up to the page with a GTM container. 

In this case, you need to set up the custom event in GTM, so GTM knows to send the event on to your GA4 property. See [Setting up custom Google Analytics events](https://newjersey.github.io/innovation-engineering/guides/monitoring/setting-up-google-analytics/#setting-up-custom-google-analytics-events) in the wiki for how to do this. 

In the GTM case, the event name would be `Clicked initial button`, the Event Category dimension would be `Page feedback`, and the rating data (`Yes` or `No`) would be under the dimension `event_label`. 

Hope this helps, and let me know if you want to hop on a call to discuss further. (I struggle with communicating the GA4 vs. GTM distinction 🫠)

jachan: Incredible! Thank you for this explanation. @natekep can we update our metrics guide to include how we'd track this in GA?

jachan:
Merging now - if the response rate is very high we shall push followup to address @ezhangy 's SAND

Issue #116, opened 2026-06-24
Show First Payments Made for PTR and ANCHOR


emtwelp:
**User story:** As a PAS-1 applicant, I want to know whether a PTR or ANCHOR payment has been made (check or a direct deposit) related to my 2025 application, so I know whether to watch for it, and if it doesn't arrive, I know to reach out.

For context: we are not including Stay NJ in this work because: 1) we don't expect any Stay NJ checks to be selected until Feb 2027, and 2) this requires showing multiple checks per program.

**Requirements**
* If a resident has one or more 2025 transactions per program with trans_x_cde = 'RF', use the first such transaction (according to check_x_dte) to show that a payment was made on the results page according to [designs](https://www.figma.com/design/fc59D8BV14x7EhW79ued5f/PAS-1-status-checker?node-id=829-10439&t=Ehj0Jm1jEtGObLRF-4).
* **FYI - we can't release this feature until Jul 15 because while some residents will have 'RF' in trans_x_cde before then, payments won't actually be made until Jul 15.** 
* Note: if the second and third characters in check_x_num = 'NN' then the payment is a direct deposit (and content should say "direct deposit made on check_x_dte" v. "check sent on check_x_dte").

emtwelp:
Including testing SSN and zip combos from TAXU table. These are fake residents with PTR transactions. 
- 020383155, 08884
- 049461541, 08049
- 150987817, 08852

PR #123, opened 2026-06-30
Prefactor 116 - Switch from URL params to using React State

jachan:
## Link to ticket

This commit is a pre-factor before we implement https://github.com/newjersey/tax-relief-status-checker/issues/116

## LLM Disclaimer
An LLM created the first draft of this code, I worked on making it more readable and removing unnecessary types/code.

## What was done?

- Previously, we put state in the URL as params, this commit moves us to using React State with a [ContextProvider](https://react.dev/learn/passing-data-deeply-with-context), since future commits will fetch more complex data from the API.

## Accessibility
N/A, no visual changes

## Steps to test

- Run Cypress tests to make sure there is no regression to the user experience

N/A, no visual changes

## Notes
N/A

### DIFF
--- a/webapp/app/status/page.tsx
+++ b/webapp/app/status/page.tsx
@@ -15,7 +15,13 @@ const StatusPage = () => {
     }
   }, [dataStore, router]);
 
-  const { lastFourSsnDigits, zipCode, applicationDate } = dataStore!;
+  // Next.js prerenders client components during the build,
+  // returning null here allows it to render only client-side
+  if (!dataStore) {
+    return null;
+  }
+
+  const { lastFourSsnDigits, zipCode, applicationDate } = dataStore;
 
   return (
     <main id="main-content">
--- a/webapp/app/status/page.tsx
+++ b/webapp/app/status/page.tsx
@@ -1,23 +1,21 @@
-import { redirect } from "next/navigation";
+"use client";
+
+import { useEffect } from "react";
+import { useRouter } from "next/navigation";
+import { useDataStore } from "@/components/TaxReliefDataProvider";
 import Link from "next/link";
 
-interface StatusPageProps {
-  readonly searchParams: Promise<{
-    readonly lastFourSsnDigits?: string;
-    readonly zipCode?: string;
-    readonly date?: string;
-  }>;
-}
+const StatusPage = () => {
+  const router = useRouter();
+  const { dataStore } = useDataStore();
 
-const StatusPage = async ({ searchParams }: StatusPageProps) => {
-  const params = await searchParams;
-  const lastFourSsnDigits = params.lastFourSsnDigits;
-  const zipCode = params.zipCode;
-  const applicationDate = params.date;
+  useEffect(() => {
+    if (!dataStore) {
+      router.replace("/");
+    }
+  }, [dataStore, router]);
 
-  if (!lastFourSsnDigits || !applicationDate || !zipCode) {
-    redirect("/");
-  }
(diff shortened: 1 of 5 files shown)


webapp/components/TaxReliefDataProvider.tsx
```diff
@@ -0,0 +1,31 @@
+"use client";
+
+import { createContext, useContext, useState } from "react";
+import type { ReactNode } from "react";
+
+export interface TaxReliefStatusData {
+  readonly lastFourSsnDigits: string;
+  readonly zipCode: string;
+  readonly applicationDate: string;
```
mluedke2: [dust/optional] - maybe `applicationDateString` instead of `applicationDate` to represent that it's a string (not Date object) without having to check its type? but also perfectly fine without I think too

webapp/components/TaxReliefDataProvider.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+const TaxReliefData = createContext<{
+  dataStore: TaxReliefStatusData | null;
+  setDataStore: (data: TaxReliefStatusData | null) => void;
```
mluedke2, line 14: [dust] just double-checking since this just has the name `dataStore` that it's the only intended one so you don't have to rename to be more specific later. if so, perfect

webapp/components/TaxReliefDataProvider.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { createContext, useContext, useState } from "react";
+import type { ReactNode } from "react";
+
+export interface TaxReliefStatusData {
+  readonly lastFourSsnDigits: string;
+  readonly zipCode: string;
+  readonly applicationDate: string;
+}
+
+const TaxReliefData = createContext<{
```
mluedke2: [pebble] - i think this would be easier to follow if it were named `TaxReliefDataContext` -- the `Context` at the end seems to be [the norm in the docs](https://react.dev/reference/react/createContext)

webapp/components/TaxReliefDataProvider.tsx
```diff
@@ -0,0 +1,31 @@
...
+  dataStore: TaxReliefStatusData | null;
+  setDataStore: (data: TaxReliefStatusData | null) => void;
+} | null>(null);
+
+export const TaxReliefDataProvider = ({ children }: { children: ReactNode }) => {
+  const [dataStore, setDataStore] = useState<TaxReliefStatusData | null>(null);
+
+  return (
+    <TaxReliefData.Provider value={{ dataStore, setDataStore }}>{children}</TaxReliefData.Provider>
```
mluedke2: [pebble] it looks like the `.Provider` syntax is [considered legacy now](https://react.dev/reference/react/createContext#provider) since you have React 19? `Starting in React 19, you can render <SomeContext> as a provider.`

mluedke2 (approved):

webapp/app/status/page.tsx
```diff
@@ -1,24 +1,28 @@
...
+      router.replace("/");
+    }
+  }, [dataStore, router]);
 
-  if (!lastFourSsnDigits || !applicationDate || !zipCode) {
-    redirect("/");
+  // Next.js prerenders client components during the build,
+  // returning null here allows it to render only client-side
+  if (!dataStore) {
+    return null;
```
mmazanec22, line 21: [dust] Should there be a more graceful fallback here for users on slow connections? (I'm not familiar enough with Next to know if this is a real concern in this case)

mmazanec22 (commented):
Looks good! Just one question

PR #133, opened 2026-07-07
[ASK] - Added transaction level information to api backend

natekep:
## Link to ticket

This commit resolves a part of [116](https://github.com/newjersey/tax-relief-status-checker/issues/116), but not the entire issue

## LLM Disclaimer

- No LLM was used to write this code. 

## What was done?

- Added transaction-level information to the api response, in order to display checks cut to the user. 
- A `transaction` is created every time there may be money that needs to be paid to a resident. Each program (ANCHOR/PTR/StayNJ) can either have no transaction or more than one transaction.
- A `transaction` should have the following information: Status (Processing, Issue Flagged, Approved, or Payment Sent). If the transaction status is 'Issue Flagged' the transaction will also have a review category code. If the transaction status is 'Payment Sent', then the transaction will also include Payment Details. 
- Payment Details should have the payment amount, payment date, check number, and payment method (check vs direct deposit) 
- We designed the API like this because we want to be able to show user's any checks they have already been cut, which is transactions where the status is 'Payment Sent'. The frontend will take in the API response and display to the user the program name, the date the Check or Direct Deposit was sent, and the amount the user received. 
## Accessibility

- No accessibility changes

## Steps to test
- Unit tests: `index.test.ts` and `transactions.test.ts` within the lambda directory, use `npx vitest lambda/`
- Deployed API changes: We manually deployed these changed to Dev and tested them with test SSN/zip pairs to confirm API response. Access the [dev site](https://main.d1fv5gc7ba4m6a.amplifyapp.com/) and enter in the following pairs (non-prod) to confirm that the API hits database and returns the right data:
- 020383155, 08884
- 049461541, 08049
- 150987817, 08852
DM @natekep or @jachan on slack for site credentials.

No visual changes

## Notes

-The API version is 1.1 because we are hoping for it to be backwards compatible with our v1 tax status checker. We are expecting that if the user has transaction information they should be able to see it after July 15th, but if the user is still waiting for their status to be "payment_sent" then they should just see the original "Your application was received on..." page from our v1. 
-The next PR will be changing the frontend to actually consume the information from the lambda function, i.e. displaying the transaction information.


lambda/src/transaction.ts
```diff
@@ -0,0 +1,95 @@
...
+export const buildTransaction = (
+  TRANS_CDE: string,
+  TRANS_STATUS_CDE: string,
+  REVIEW_CATEGORY_CDE: string,
+  CHECK_DTE: string,
+  CHECK_AMT: number,
+  CHECK_NUM: string,
+): Transaction => {
+  let status;
+  if (TRANS_CDE == "RR" && TRANS_STATUS_CDE == "PR" && REVIEW_CATEGORY_CDE === null) {
```
timwright12: Are these `==` intentional vs the `===`? Noting about 9 instances

lambda/src/transaction.ts
```diff
@@ -0,0 +1,95 @@
...
+      row[`CHECK_${i}_NUM`] as string,
+    );
+    if (row[`TRANS_${i}_TAX_CDE`] == anchorCDE) {
+      anchor.push(transaction);
+    } else if (row[`TRANS_${i}_TAX_CDE`] == ptrCDE) {
+      ptr.push(transaction);
+    } else if (row[`TRANS_${i}_TAX_CDE`] == stayCDE) {
+      stay_nj.push(transaction);
+    } else {
+      throw new Error("No Transaction Tax Code");
```
timwright12: Good instinct adding error handling for invalid tax codes! Let's make the error message more helpful for debugging. Instead of "No Transaction Tax Code", include details about what went wrong if we can:

```throw new Error(
     Invalid transaction tax code: ${row[\`TRANS_${i}_TAX_CDE\`]} for transaction ${i}`
  );```

File: lambda/src/transactions.test.ts [hunk 3]
```diff
@@ -0,0 +1,265 @@
...
+  default: {
+    getConnection: vi.fn(() =>
+      Promise.resolve({
+        execute: mockExecute,
+        close: mockClose,
+      }),
+    ),
+    OUT_FORMAT_OBJECT: 4001,
+  },
+}));
```
aloverso: [pebble] Is this mocking necessary? I might be missing something, but it looks like the transaction code is stateless & independent from the database, because it just takes an `InquiryRow` as a parameter. (Which I think is a good thing!)

Because of that, it seems like these tests should not need to know anything about the database or the secrets manager. I think you could also remove the `mockExecute.mockResolvedValue({ rows: [row] });` line from each test? Also the setup in the beforeEach

lambda/src/transaction.ts
```diff
@@ -0,0 +1,95 @@
...
+
+/** Database row from ELF_SAVER_INQUIRY */
+export interface InquiryRow {
+  readonly SOCIAL_SECURITY_NUMBER_IDN: string;
+  readonly ZIP_ADR: string;
+  readonly RNY_APPLIED_DTE: string;
+  readonly RETURN_YEAR_DTE: number;
+  readonly TRANS_TOTAL_NUM: number;
+  readonly [key: string]: unknown;
+}
```
aloverso: [pebble] I see this interface is already defined in `index.ts` - I recommend not rewriting/redefining it here. 

It's a fun conundrum. You need access to this type, which is currently in `index` and not exported, so cannot be imported here from index. You can't add an export to the type in index, because index imports from this `transaction` file, so having this file also import from `index` would create a circular dependency.

The solution I would recommend is to create a new `types.ts` file that exports all the types necessary, and that both `index` and `transactions` can import from. You could include also the `Transaction` and `AllTransactions` types as well

lambda/src/transaction.ts
```diff
@@ -0,0 +1,95 @@
...
+
+const anchorCDE = 13;
+const ptrCDE = 49;
+const stayCDE = 41;
```
aloverso: [sand] These are fixed consts, right? Using all uppercase variables names would help indicate that these are fixed values, not variables.

lambda/src/transaction.ts
```diff
@@ -0,0 +1,95 @@
...
+  for (let i = 1; i <= row.TRANS_TOTAL_NUM; i++) {
+    const transaction = buildTransaction(
+      row[`TRANS_${i}_CDE`] as string,
+      row[`TRANS_STATUS_${i}_CDE`] as string,
+      row[`REVIEW_CATEGORY_${i}_CDE`] as string,
+      row[`CHECK_${i}_DTE`] as string,
+      row[`CHECK_${i}_AMT`] as number,
+      row[`CHECK_${i}_NUM`] as string,
+    );
+    if (row[`TRANS_${i}_TAX_CDE`] == anchorCDE) {
```
aloverso: [dust] The `row[TRANS_${i}_TAX_CDE]` is used repeatedly, could extract it into a variable for more readable code

lambda/src/transactions.test.ts
```diff
@@ -0,0 +1,265 @@
...
+      const row = buildMockRow({
+        TRANS_1_CDE: "RR",
+        TRANS_STATUS_1_CDE: "APR",
+      });
+      mockExecute.mockResolvedValue({ rows: [row] });
+
+      const result = callBuildTransaction(row, 1);
+      expect(result.status).toBe("approved");
+      expect(result.payment_details).toBeUndefined();
+    });
```
aloverso: [dust] Why have two separate tests here? It seems like the logic is identical, and the code does not make a distinction between APC and APR, right?

aloverso: It seems like you want to test that this works the same regardless of whether it is APC or APR - that makes sense! I think you could reformat the tests to take in a parameter, using `it.each`. This helps because then there is only one test body and output, which helps visually demonstrate that the two inputs have the same expected output.

This looks like:
```javascript
it.each([
  "APC",
  "APR",
  "APX",
])("returns status as Approved without payment details when status code is %s", async (statusCode) => {
    const row = buildMockRow({
        TRANS_1_CDE: "RR",
        TRANS_STATUS_1_CDE: statusCode,
      });
     
    // ...rest of your test here
});

```

lambda/src/transactions.test.ts
```diff
@@ -0,0 +1,265 @@
...
+        CHECK_1_NUM: "012345678",
+      });
+      mockExecute.mockResolvedValue({ rows: [row] });
+
+      const result = callBuildTransaction(row, 1);
+      expect(result.status).toBe("payment_sent");
+      expect(result.payment_details?.amount).toBe(1750);
+      expect(result.payment_details?.method).toBe("check");
+      expect(result.payment_details?.check_number).toBe("012345678");
+      expect(result.payment_details?.date).toBe("12/11/2025 00:00:00");
```
aloverso: [dust] Because this test (and the one below it) are _specifically_ testing the `method` output, I think you could delete the other expectations! (status, amount, number, date) 

Reasoning: 

Tests are structured like "given X, when Y, then Z". In this case, given a row with a certain CHECK_NUM, when built, then the method is "check".

A good test, imo, makes the cause-and-effect relationship between the "given" and the "then" as clear as possible. One way to do that is to ensure that what we test for in the expectations is clearly tied to the "given". So removing all the expectations that are unrelated to the CHECK_NUM (which is the behavior under test) makes for a clear cause-and-effect story that is easy to read in the code at a glance. 

This works specifically because you have the above test which confirms the correct output for all those other fields already!

lambda/src/transactions.test.ts
```diff
@@ -0,0 +1,265 @@
...
+        CHECK_5_AMT: 246,
+        CHECK_5_NUM: "514221081",
+        TRANS_5_TAX_CDE: 41,
+      });
+      mockExecute.mockResolvedValue({ rows: [row] });
+
+      const result = buildAllTransactions(row);
+      expect(result.anchor.length).toBe(2);
+      expect(result.ptr.length).toBe(1);
+      expect(result.stay_nj.length).toBe(2);
```
aloverso, line 212: [dust] It might be nice to add a couple extra expectations here to give extra confidence. As in, specifically looking at a field in the `result.anchor[0]` and `result.anchor[1]` and `result.ptr[0]` etc objects, and confirm that the right transaction was sorted into the right bucket.

lambda/src/index.test.ts
```diff
@@ -148,6 +148,9 @@ describe("handler business logic", () => {
       expect(body["records"]).toHaveLength(1);
       expect(body["records"][0].return_year).toBe("2024");
       expect(body["records"][0].application_date).toBe("10/31/2025 00:00:00");
+      expect(body["records"][0].anchor).toBeDefined();
+      expect(body["records"][0].ptr).toBeDefined();
+      expect(body["records"][0].stay_nj).toBeDefined();
```
aloverso: [pebble] Could we make this test more specific and confirm that not only are they defined, the specific values match what we expect from the data in the database row?

lambda/src/transaction.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  if (status === "issue_flagged") {
+    return { status: status, review_category: REVIEW_CATEGORY_CDE };
+  }
+
+  if (status === "payment_sent") {
+    let method = "check";
+    if (CHECK_NUM.slice(1, 3) === "NN") {
+      method = "direct_deposit";
+    }
```
aloverso: [dust] Is it possible that CHECK_NUM might not be defined? Looking at this and line 48, I see that there's a world where these could be undefined but it returns "payment_sent" via "check" since they are both `else` conditions. Is that desired?

natekep: I'm looking through the TAXU database test records and for all the transactions where the status is "payment_sent" which is the only ones where there should be payment details, each transaction has a CHECK_NUM defined. I am wondering then should I still check for and throw an error if CHECK_NUM is undefined just in case?

lambda/src/index.test.ts
natekep: We test the logic for the data row information more in depth within transaction.test.ts, so I am wondering if you think its still worth testing for here

lambda/src/transaction.ts
jachan: My 2c, just because the database *should* have CHECK_NUM defined doesn't mean we can rely on it to always have it defined. Better to throw an error if we encounter something unexpected.

lambda/src/transaction.ts
```diff
@@ -0,0 +1,97 @@
...
+      date: CHECK_DTE,
+      method: method,
+      check_number: CHECK_NUM,
+    };
+    return { status: status, payment_details: payment_details };
+  }
+  return { status: status };
+};
+
+export const buildAllTransactions = (row: InquiryRow): AllTransactions => {
```
jachan, line 50: [DUST] Really like where we ended up with this function, good job!

lambda/src/transactions.test.ts
```diff
@@ -0,0 +1,265 @@
...
+  RNY_APPLIED_DTE: "10/31/2025 00:00:00",
+  TRANS_TOTAL_NUM: 2,
+  TRANS_1_CDE: "RF",
+  TRANS_STATUS_1_CDE: "APC",
+  REVIEW_CATEGORY_1_CDE: "MDZ",
+  CHECK_1_DTE: "12/11/2025 00:00:00",
+  CHECK_1_AMT: 1750,
+  CHECK_1_NUM: "922775385",
+  TRANS_1_TAX_CDE: 13,
+  TRANS_2_CDE: "RF",
```
jachan: [DUST] Is the 2nd transaction actually used in any of the tests other than "when there are MANY transactions"? If not, prefer moving those fields to that test's specific setup. This should help readability, as it might be confusing to someone encountering the test for the first time why these fields are part of a default Mock Row

lambda/src/transactions.test.ts
```diff
@@ -0,0 +1,265 @@
...
+  TRANS_STATUS_2_CDE: "APC",
+  REVIEW_CATEGORY_2_CDE: null,
+  CHECK_2_DTE: "11/28/2025 0:00:00",
+  CHECK_2_AMT: 246,
+  CHECK_2_NUM: "514221081",
+  TRANS_2_TAX_CDE: 49,
+  DIRECT_DEPOSIT_IND: "Y",
```
jachan: [PEBBLE] Suggest removing this field as we're no longer relying on DIRECT_DEPOSIT_IND in our logic

jachan (approved):
Apart from Anne and Tim's comments, left a few other thoughts. Great work here!

PR #144, opened 2026-07-13
[ASK] - Show first payments made for PTR and ANCHOR

natekep:
## Link to ticket

This commit resolves a part of [116](https://github.com/newjersey/tax-relief-status-checker/issues/116), having the frontend consume the api response from this [previous PR](https://github.com/newjersey/tax-relief-status-checker/pull/137)

## LLM Disclaimer

## What was done?

- Created a separate `app/payment/` page to display to a whether a PTR or ANCHOR payment has been made (check or a direct deposit) related to a user's 2025 application. 
- The page will only display the first transaction per category where the status is `payment_sent`, and that ordering is based on a payment's `CHECK_DTE`. 
- Added routing to home page, where if a user's record has a transaction with status = `payment_sent`, and if feature flag = true, then the user gets pushed to the page displaying their payments, otherwise it goes to the original "Your application was received..." page.
- The feature flag is needed because until Jul 15 some residents will have a transaction with `payment_sent` before then, but payments won't actually be made until Jul 15. See Notes for more information about this. 
- Changed CI to include env variable for feature flag for e2e testing

## Accessibility

## Steps to test

- For local testing, start the server with `NEXT_PUBLIC_ENABLE_SHOW_PAYMENTS="true" npm run dev` for cypress tests of the showPayments new feature.

- Before
- After

## Notes

Because the feature flag is going to be set to true on 8/15, we only created one version of tests where the flag is true, instead of testing two parallel versions of app+cypress test (each with a differently .env-configured application process and a different corresponding cypress assertion) due to that time constraint.


.github/workflows/end_to_end_tests.yml
```diff
@@ -21,5 +21,5 @@ jobs:
           working-directory: webapp
           install: false
           build: npx next build
-          start: npx next start
+          start: NEXT_PUBLIC_ENABLE_SHOW_PAYMENTS="true" npx next start
```
jachan: [DUST] Is the plan to leave this forever, or make a change after July 15th to remove this feature flag?

webapp/app/payment-info/page.tsx
jachan, line 1: [PEBBLE] Can we change the name of this page to be something more meaningful? `payment-info`, perhaps?

webapp/app/payment-info/page.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+import { useEffect } from "react";
+import { useRouter } from "next/navigation";
+import { useDataStore } from "@/components/TaxReliefDataProvider";
+import Link from "next/link";
+import { Table } from "@trussworks/react-uswds";
+import { formatDate } from "../utils/formatDate";
+import { expandFaqAccordionItem, LandingPageFaq } from "@/components/LandingPageFaq";
+
+const getEarliestTransaction = (transactions: any[]) => {
```
jachan: [PEBBLE] Don't we have a `Transaction` type? Can we use it here instead of `any` to give us better type-safety?

webapp/app/payment/page.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const valid = transactions.filter((t) => t.status === "payment_sent" && t.payment_details?.date);
+  if (valid.length === 0) return;
+  return valid.reduce((earliest, current) =>
+    new Date(current.payment_details.date) < new Date(earliest.payment_details.date)
+      ? current
+      : earliest,
+  );
+};
+
+const PaymentPage = () => {
```
jachan: [PEBBLE] If you rename the page path, you'll also want to rename this.

webapp/app/payment-info/page.tsx
```diff
@@ -0,0 +1,137 @@
...
+                    )}
+                    <td>${transaction.payment_details.amount}</td>
+                  </tr>
+                );
+              })()}
+            </tbody>
+          </Table>
+          <div className="grid-row grid-gap margin-top-5">
+            <h2 className="font-heading-l">Frequently Asked Questions (FAQs)</h2>
+            <LandingPageFaq headingLevel="h3" />
```
jachan: [PEBBLE] These are no longer on the `LandingPage`, so this name is inaccurate. Which FAQs should be shown on this page? If we envision a different set of FAQs on each page, then I can see an argument for making a new component for the `PaymentInfoPageFAQs`. If we envision the same set of FAQs on every page, then I'd rename the component to something like `PropertyTaxReliefFAQs`.

webapp/app/page.tsx
```diff
@@ -119,10 +134,17 @@ const LandingPage = () => {
...
+        ptr: record2025.ptr,
+        stay_nj: record2025.stay_nj,
       });
 
       logGAEvent(`api_200_record_found`);
-      router.push("/status");
+      if (process.env.NEXT_PUBLIC_ENABLE_SHOW_PAYMENTS == "true" && checkPaymentsSent(record2025)) {
+        router.push("/payment");
+      } else {
+        router.push("/status");
```
jachan: [PEBBLE] Along with renaming the `payment` page to `payment-info`, I suggest renaming the `status` page to `application-received`.

webapp/components/LandingPageFaq.tsx
```diff
@@ -1,3 +1,5 @@
+"use client";
```
jachan: [DUST] Does this need to be a client component? Why?

webapp/cypress/e2e/paymentInfo.cy.ts
```diff
@@ -0,0 +1,231 @@
...
+  cy.contains("th", "Program").should("be.visible");
+  cy.contains("th", "Payment Status").should("be.visible");
+  cy.contains("th", "Amount").should("be.visible");
+
+  cy.contains("td", "ANCHOR").should("be.visible");
+  cy.contains("td", `Direct deposit made on ${mockDate}`).should("be.visible");
+  cy.contains("td", `$${mockAmount}`).should("be.visible");
+});
+
+it("should not display stay_nj CHECK", () => {
```
jachan, line 137: Context for other reviewers: we've been asked to implement `stay_nj` in a future ticket, not in this PR.

webapp/cypress/e2e/applicationFound.cy.ts
jachan, line 1: [DUST] What is the distinction between the tests in `userPath` and `showPayments`? Suggest renaming this file to be more specific.

webapp/cypress/e2e/applicationFound.cy.ts
```diff
@@ -128,9 +128,77 @@ it("should display api alert if records is empty in a 200 response", () => {
 
 it("should display status page if records has an object in a 200 response", () => {
   fillFields();
-  cy.intercept("POST", "/api/status", {
-    statusCode: 200,
-    fixture: "v1_api_found_records.json",
```
jachan, line 133: [PEBBLE] If v1 of the API is no longer present, suggest removing the v1 files from the repo in this commit. They'll still be in the version history if we need to reference them in the future.

webapp/cypress/e2e/userPath.cy.ts
```diff
@@ -128,9 +128,77 @@ it("should display api alert if records is empty in a 200 response", () => {
 
 it("should display status page if records has an object in a 200 response", () => {
   fillFields();
-  cy.intercept("POST", "/api/status", {
-    statusCode: 200,
-    fixture: "v1_api_found_records.json",
+  cy.fixture("v2_api_found_records.json").then((resp) => {
```
jachan: [DUST] This seems weird to me because we're taking a fixture that's called `found_records` and mutating it so that no records are found. In this case, I'd prefer to just create a fixture called `v2_api_no_records.json`. I think that's readable, but leaving it up to you.

webapp/cypress/e2e/applicationFound.cy.ts
```diff
@@ -128,9 +128,77 @@ it("should display api alert if records is empty in a 200 response", () => {
...
+    "event",
+    "api_200_record_found",
+    Cypress.sinon.match.any,
+  );
+
+  cy.contains("a", "Log out").click();
+  cy.contains("h1", "This website is checking your 2025 PAS");
+});
+
+it("should display application found page if records has an object, but no transactions are payment_sent", () => {
```
jachan, line 160: [DUST] Love this test!

webapp/cypress/e2e/userPath.cy.ts
```diff
@@ -128,9 +128,77 @@ it("should display api alert if records is empty in a 200 response", () => {
...
+    "event",
+    "api_200_record_found",
+    Cypress.sinon.match.any,
+  );
+
+  cy.contains("a", "Log out").click();
+  cy.contains("h1", "This website is checking your 2025 PAS");
+});
+
+it("should display application found page if records has an object with payments sent, but feature flag is false", () => {
```
jachan: [DUST] How are you setting the feature flag to false here?

jachan (commented):
Left some questions and some naming suggestions, but the logic and testing looks solid!

webapp/app/payment-info/page.tsx
natekep: Do you know what we envision for the future? Or is this a question I should ask Amy

jachan: I'm not sure, you'd have to ask Amy!

.github/workflows/end_to_end_tests.yml
natekep: Plan to remove feature flag after July 15th

webapp/cypress/e2e/userPath.cy.ts
natekep: This test was passing because there were no transactions with payment sent, so it was not testing the right functionality. We are unable to set feature flag to false based on how we set the env variable in CI for e2e testing. We decided to remove this test to only test functionality when the feature flag is true.

webapp/app/payment-info/page.tsx
```diff
@@ -0,0 +1,147 @@
...
+                <use href="/img/sprite.svg#logout"></use>
+              </svg>
+              Log out
+            </Link>
+          </div>
+          <div>
+            <div className="grid-row">
+              <div className="tablet:grid-col-3">
+                <p>
+                  SSN/ITIN: <strong>***-**-{lastFourSsnDigits}</strong>
```
jachan, line 65: [SAND] This header is shared between the `payment-info` and `application-received` pages, suggest future refactor to dry up this code.

webapp/app/payment-info/page.tsx
```diff
@@ -0,0 +1,147 @@
...
+  const { lastFourSsnDigits, zipCode, anchor, ptr } = dataStore;
+
+  return (
+    <main id="main-content">
+      <section className="usa-section">
+        <div className="grid-container">
+          <div style={{ textAlign: "right" }}>
+            <Link className="usa-button usa-button--outline margin-right-3 margin-top-3" href="/">
+              <svg focusable="false" role="img" width="20" height="20" fill="#005ea2">
+                <use href="/img/sprite.svg#logout"></use>
```
jachan, line 56: [SAND] This does not match the designs, since logout button is below the Property Tax Relief Status Checker and logo, when it should be above it. 

We can fix this as a bug-fix later, I believe it's been around for a while.

webapp/app/page.tsx
```diff
@@ -22,8 +22,23 @@ interface UserData {
 interface StatusRecord {
   readonly return_year: string;
   readonly application_date: string;
+  readonly anchor: [];
+  readonly ptr: [];
+  readonly stay_nj: [];
 }
-
+const checkPaymentsSent = (record: StatusRecord) => {
```
jachan: [DUST] Prefer `hasPaymentSentTransaction` but flexible on naming

webapp/cypress/e2e/paymentInfo.cy.ts
jachan, line 1: [DUST] Consider renaming this file `paymentInfo.cy.ts` to match new page

webapp/components/TaxReliefDataProvider.tsx
```diff
@@ -3,10 +3,23 @@
 import { createContext, useContext, useState } from "react";
 import type { ReactNode } from "react";
 
+interface Transaction {
```
jachan: [PEBBLE] This seems duplicated with `types.d.ts` below.

jachan (approved):
Approved once these comments are addressed!

PR #145, opened 2026-07-15
[ASK] Add API StatusCode metrics

jachan:
## Link to ticket

No ticket exists

## LLM Disclaimer

- An LLM generated the first draft of this PR, which was reviewed and revised by me manually for clarity and readability

## What was done?

- This ticket takes advantage of the [Amazon CloudWatch EMF (embedded metric format)](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Specification.html) to publish CloudWatch Metrics using JavaScript `console.log`.
- Added alarm to CDK
- The goal is to alarm when the API has a 500 error.

## Accessibility
N/A, this is a metrics change

## Steps to test

- This was deployed to dev by running `npm run deploy:dev` and I tested by calling the Lambda directly with both a properly formatted SSN (200) and a improperly formatted input (400). Verified that we were able to see the metrics appear instantly in Cloudwatch (see screenshot below)

image

### DIFF
--- a/infra/lib/infra-stack.ts
+++ b/infra/lib/infra-stack.ts
@@ -41,7 +41,7 @@ export class InfraStack extends Stack {
       alarmName: "PropertyTaxReliefStatusApi500Alarm",
       alarmDescription: "Alarm for API errors",
       comparisonOperator: cw.ComparisonOperator.GREATER_THAN_THRESHOLD,
-      treatMissingData: cw.TreatMissingData.MISSING,
+      treatMissingData: cw.TreatMissingData.NOT_BREACHING,
     });
 
     const vpc = ec2.Vpc.fromLookup(this, `vpc-${props.stageName}`, {
--- a/infra/lib/infra-stack.ts
+++ b/infra/lib/infra-stack.ts
@@ -6,6 +6,7 @@ import * as ec2 from "aws-cdk-lib/aws-ec2";
 import * as iam from "aws-cdk-lib/aws-iam";
 import * as path from "path";
 import { fileURLToPath } from "url";
+import { Alarm, ComparisonOperator, Metric, TreatMissingData } from "aws-cdk-lib/aws-cloudwatch";
 
 const DIRNAME = path.dirname(fileURLToPath(import.meta.url));
 const LAMBDA_TIMEOUT_SECONDS = 5;
@@ -26,6 +27,22 @@ export class InfraStack extends Stack {
 
   constructor(scope: Construct, id: string, props: InfraStackProps) {
     super(scope, id, props);
+
+    new Alarm(this, "PropertyTaxReliefStatusApi500Alarm", {
+      metric: new Metric({
+        namespace: "TaxReliefStatusApi",
+        metricName: "ResponseCount",
+        dimensionsMap: { statusCode: "500" },
+        statistic: "Sum",
+      }),
+      threshold: 0,
+      evaluationPeriods: 1,
+      datapointsToAlarm: 1,
+      alarmName: "PropertyTaxReliefStatusApi500Alarm",
+      alarmDescription: "Alarm for API errors",
+      comparisonOperator: ComparisonOperator.GREATER_THAN_THRESHOLD,
+      treatMissingData: TreatMissingData.MISSING,
+    });
 
     const vpc = ec2.Vpc.fromLookup(this, `vpc-${props.stageName}`, {
       vpcId: props.vpcId,
--- a/infra/lib/infra-stack.ts
+++ b/infra/lib/infra-stack.ts
@@ -6,7 +6,7 @@ import * as ec2 from "aws-cdk-lib/aws-ec2";
 import * as iam from "aws-cdk-lib/aws-iam";
 import * as path from "path";
 import { fileURLToPath } from "url";
-import { Alarm, ComparisonOperator, Metric, TreatMissingData } from "aws-cdk-lib/aws-cloudwatch";
+import * as cw from "aws-cdk-lib/aws-cloudwatch";
 
 const DIRNAME = path.dirname(fileURLToPath(import.meta.url));
 const LAMBDA_TIMEOUT_SECONDS = 5;
@@ -28,8 +28,8 @@ export class InfraStack extends Stack {
   constructor(scope: Construct, id: string, props: InfraStackProps) {
     super(scope, id, props);
 
-    new Alarm(this, "PropertyTaxReliefStatusApi500Alarm", {
-      metric: new Metric({
+    new cw.Alarm(this, "PropertyTaxReliefStatusApi500Alarm", {
+      metric: new cw.Metric({
         namespace: "TaxReliefStatusApi",
         metricName: "ResponseCount",
         dimensionsMap: { statusCode: "500" },
@@ -40,8 +40,8 @@ export class InfraStack extends Stack {
       datapointsToAlarm: 1,
       alarmName: "PropertyTaxReliefStatusApi500Alarm",
       alarmDescription: "Alarm for API errors",
-      comparisonOperator: ComparisonOperator.GREATER_THAN_THRESHOLD,
-      treatMissingData: TreatMissingData.MISSING,
+      comparisonOperator: cw.ComparisonOperator.GREATER_THAN_THRESHOLD,
+      treatMissingData: cw.TreatMissingData.MISSING,
     });
 
     const vpc = ec2.Vpc.fromLookup(this, `vpc-${props.stageName}`, {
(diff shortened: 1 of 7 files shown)


jviall (commented):
This is a really cool feature of cloudwatch! But I wonder if it's the right way to capture 4xx/5xx errors? Are ya'll utilizing API gateway? if so [it has built in metrics](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-metrics-and-dimensions.html) for status codes that will capture more than just 400/500 errors (e.g. 503), and seems like the more thorough way to implement this. But I forget if in your system design you ended up utilizing API gateway.

jachan:
We are not currently using API Gateway, so we do not have the built-in metrics from that! 

We are using a Lambda function URL instead for now.

jachan:
Ended up rolling the CDK Alarm change into this as well.

lambda/src/index.test.ts
```diff
@@ -79,17 +83,26 @@ describe("handler input validation", () => {
...
     mockExecute.mockRejectedValue(new Error("ORA-12541: TNS:no listener"));
+    const consoleLogSpy = vi.spyOn(console, "log").mockImplementation(() => {});
 
     const result = await handler({ ssn: "123456789", zip: "07656" });
 
     expect(result.statusCode).toBe(500);
     expect(JSON.parse(result.body).error).toBe("Internal server error");
+    expect(
+      consoleLogSpy.mock.calls.some((call) => String(call[0]).includes('"StatusCode":"500"')),
+    ).toBe(true);
```
sannidhishukla: [dust] is there a reason for checking that `console.log` is called with the expected params here rather than checking that `embeddedMetricsPublisher.recordResponse` is called with the expected params? this feels a bit like testing the implementation specifics for `embeddedMetricsPublisher.recordResponse` rather than just testing the `handler` function

infra/lib/infra-stack.ts
```diff
@@ -27,6 +28,22 @@ export class InfraStack extends Stack {
   constructor(scope: Construct, id: string, props: InfraStackProps) {
     super(scope, id, props);
 
+    new Alarm(this, "PropertyTaxReliefStatusApi500Alarm", {
```
sannidhishukla: [question] do you plan to set up an alarm for status code 400? and if not how come, out of curiosity?

sannidhishukla (commented):
looks great, so cool to learn about EMF!! left one small comment + a question

infra/lib/infra-stack.ts
```diff
@@ -6,6 +6,7 @@ import * as ec2 from "aws-cdk-lib/aws-ec2";
 import * as iam from "aws-cdk-lib/aws-iam";
 import * as path from "path";
 import { fileURLToPath } from "url";
+import { Alarm, ComparisonOperator, Metric, TreatMissingData } from "aws-cdk-lib/aws-cloudwatch";
```
casewalker: dust/sand - The previous style has been to import the entire cdk-lib subdir under a short name, to be inkeeping with the previous style, I would suggest doing the same here:
```suggestion
import * as cw from "aws-cdk-lib/aws-cloudwatch";
```

infra/lib/infra-stack.ts
casewalker: I would guess 400 represents user-error whereas 500 is a server error, so is probably worth knowing about, but might not be worth alerting on?

lambda/src/index.test.ts
```diff
@@ -1,4 +1,4 @@
-import { describe, it, expect, beforeEach, vi } from "vitest";
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
```
casewalker, line 1: dust - out-of-scope, shouldn't vitest be configured so that these globals are imported globally and don't need to be explicitly imported?

infra/lib/infra-stack.ts
jachan: I originally planned to alarm on both 400 and 500 errors, but since we're using 400 to capture client side errors (e.g. malformatted input: https://github.com/newjersey/tax-relief-status-checker/blob/main/lambda/src/index.ts#L111), it feels like we wouldn't want to alarm on it, as it's something we expect to happen.

lambda/src/index.test.ts
```diff
@@ -34,6 +34,10 @@ beforeEach(() => {
   });
 });
 
+afterEach(() => {
+  vi.restoreAllMocks();
+});
```
casewalker, line 40: dust - I have fun rewriting these as:
```suggestion
afterEach(vi.restoreAllMocks);
```

I am 90% sure it works.

lambda/src/index.test.ts
casewalker: dust/sand - This looks pretty unreadable to me, if you do keep it rather than going in the direction Sanni suggests, I would change it to:
```suggestion
expect(consoleLogSpy).toHaveBeenCalledWith(expect.stringMatching('"StatusCode":"500"'));
```

lambda/src/index.test.ts
```diff
@@ -105,12 +118,16 @@ describe("handler business logic", () => {
...
+      const consoleLogSpy = vi.spyOn(console, "log").mockImplementation(() => {});
 
       const result = await handler({ ssn: "123456789", zip: "00000" });
 
       expect(result.statusCode).toBe(200);
       const body = JSON.parse(result.body);
       expect(body.records).toEqual([]);
+      expect(
+        consoleLogSpy.mock.calls.some((call) => String(call[0]).includes('"StatusCode":"200"')),
+      ).toBe(true);
```
casewalker: dust/sand - Same comment as before:

```suggestion
expect(consoleLogSpy).toHaveBeenCalledWith(expect.stringMatching('"StatusCode":"200"'));
```

lambda/src/metrics.ts
```diff
@@ -0,0 +1,59 @@
+/** CloudWatch namespace under which all custom metrics emitted by this API are published. */
+const METRICS_NAMESPACE = "TaxReliefStatusApi";
```
casewalker, line 2: dust - Just be aware and make sure you are happy with this namespace and the metric names, I think @jviall played with custom metrics in CCM awhile back, and once you publish metrics I think they live for like 15 months?

infra/lib/infra-stack.ts
sannidhishukla: ooh got it, makes sense. thanks!!

lambda/src/metrics.ts
```diff
@@ -0,0 +1,59 @@
...
+export interface MetricsPublisher {
+  recordResponse(params: RecordResponseParams): void;
+}
+
+/**
+ * Shape of a single CloudWatch Embedded Metric Format (EMF) log entry. CloudWatch Logs' backend
+ * log-processing pipeline automatically parses any log line matching this envelope into a real
+ * CloudWatch custom metric.
+ */
+interface EmfLogEntry {
```
casewalker, line 20: sand - I think the "right way" to do this would be to add the dependency [aws-embedded-metrics](https://github.com/awslabs/aws-embedded-metrics-node/) and then use it to construct and output EMF logs. It takes away a lot of your boilerplate here, it is helpful if you ever want to start logging more complex metrics, it has some string constant exports that are a little nice... but it is probably annoying to set it up right (it has some extra stuff by default), and it's a whole extra dependency, so I guess it's up to you.

But I think it would look like this:
```
import { createMetricsLogger, Unit } from "aws-embedded-metrics";

const recordResponse = async (statusCode: number): Promise<void> => {
  const metrics = createMetricsLogger();
  metrics.setNamespace("TaxReliefStatusApi");
  metrics.resetDimensions(false);
  metrics.putDimensions({ StatusCode: String(statusCode) });
  metrics.putMetric("ResponseCount", 1, Unit.Count);
  await metrics.flush();
};
```

casewalker (approved):
Looks pretty good, left a number of comments, I am most curious for your thoughts on the aws-embedded-metrics dependency.

lambda/src/index.test.ts
jachan: No, I agree that this approach is better, thank you for the note!

lambda/src/index.test.ts
jachan: Really appreciate it. Wasn't as familiar with using `Spy`s, agree that your syntax is much more legible.

lambda/src/metrics.ts
jachan: Think I'm going to stick with this approach for now, since it feels pretty lightweight to me. Thanks for bringing my attention to the alternative, though!

casewalker: IMO it is a light lift to do it the right way, it is better long-term for NJIA projects, it is a hack to do it this way that sets a poor precedent.

PR #146, opened 2026-07-16
[SHOW] - 116 - api explicitly check for payment sent trans code 

natekep:
## Link to ticket

This commit resolves a part of [116](https://github.com/newjersey/tax-relief-status-checker/issues/116), 

## LLM Disclaimer

- No LLM was used to write code. 

## What was done?

- API now sets transaction status to `payment_sent` now iff the `TRANS_CDE === "RF"`, and throws an error otherwise. 
- This had to be done because while setting up alarms, we ran into errors in prod. The main cause appears to be that there were some transactions that were missing `CHECK_NUM`, but had a status code of `payment_sent`. This should not happen with our current design, and the previous implementation had mislabeled these transactions by defaulting the status code to `payment_sent`
- Changed missing `CHECK_NUM` error throw to be before string slice to prevent against reading on null

## Accessibility

-No accessibility changes

## Steps to test

- New unit test for when `TRANS_CDE` is missing or not one of `RR` or `RF`.

No visual changes

### DIFF
--- a/lambda/src/transaction.ts
+++ b/lambda/src/transaction.ts
@@ -19,8 +19,10 @@ export const buildTransaction = (
     status = "issue_flagged";
   } else if (TRANS_CDE === "RR" && TRANS_STATUS_CDE.startsWith("AP")) {
     status = "approved";
-  } else {
+  } else if (TRANS_CDE === "RF") {
     status = "payment_sent";
+  } else {
+    throw new Error(`Invalid TRANS_CDE: ${TRANS_CDE}`);
   }
 
   if (status === "issue_flagged") {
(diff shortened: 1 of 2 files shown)


jachan:
Hi Nate, can we also fix the CHECK_NUM `slice` bug in this PR?

lambda/src/transaction.ts
```diff
@@ -19,30 +19,34 @@ export const buildTransaction = (
...
   }
 
   if (status === "payment_sent") {
-    let method;
-    if (CHECK_NUM.slice(1, 3) === "NN") {
-      method = "direct_deposit";
-    } else if (CHECK_NUM.slice(1, 3)) {
-      method = "check";
-    } else {
+    if (!CHECK_NUM) {
```
jachan, line 33: [PEBBLE] Can we add a unit test for this case and make sure it's throwing the error correctly?

jachan (approved):

PR #148, opened 2026-07-16
[SHOW] bugfix - handle PRH transaction code

jachan:
## Link to ticket

No ticket exists for this PR

## LLM Disclaimer

No, an LLM was not used

## What was done?
- Fixed issue where a previously unknown Transaction Code would cause the API to error. Consulted with Product, and the correct logic should be `startsWith("PR")` rather than expecting Transaction Code to `=== "PR"`

## Accessibility
N/A, API change

## Steps to test
run `npm run test`, created two new unit tests for the usecase.

N/A

## Notes
None

### DIFF
--- a/lambda/src/transaction.ts
+++ b/lambda/src/transaction.ts
@@ -13,9 +13,13 @@ export const buildTransaction = (
   CHECK_NUM: string,
 ): Transaction => {
   let status;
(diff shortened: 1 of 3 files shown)


natekep (approved):
Looks Good!

Issue #153, opened 2026-07-17
[WIP] Clarify actual desired behavior moving forward for feature flags


emtwelp:

jachan:
PR above is unrelated.

EXPLORATORY: This ticket is meant to capture if we can do this without feature flags

PR #159, opened 2026-07-17
[SHOW] 153 - remove NEXT_PUBLIC_ENABLE_SHOW_PAYMENTS feature flag

jachan:
## Link to ticket

https://github.com/newjersey/tax-relief-status-checker/issues/153

## LLM Disclaimer

- No LLM was used for this PR

## What was done?

- Previously, we gated routing users to the `payment-info` page behind a feature flag. Now that the July 15th launch date has passed, we no longer need to support this flag.

## Accessibility
N/A

## Steps to test
- Ran server locally WITHOUT feature flag using `npm run test`
- Ran Cypress against local server using `npm run cypress` from project root
- Validated that Cypress tests passed. (Using version from `main`, cypress tests would fail unless feature flag was set with `NEXT_PUBLIC_ENABLE_SHOW_PAYMENTS="true" npm run dev`)

None

### DIFF
--- a/.github/workflows/end_to_end_tests.yml
+++ b/.github/workflows/end_to_end_tests.yml
@@ -23,5 +23,3 @@ jobs:
           build: npx next build
           start: npx next start
           wait-on: "http://localhost:3000/"
-        env:
-          NEXT_PUBLIC_ENABLE_SHOW_PAYMENTS: "true"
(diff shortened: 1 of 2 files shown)


natekep (approved):
Looks good!

PR #163, opened 2026-07-21
[ASK] - 143 Add update payments for PTR and ANCHOR

### LINKED ISSUE #143: Add update payments for PTR and ANCHOR
**Updated Aug 5: This ticket will only focus on update payments for PTR and ANCHOR. We'll handle Stay NJ payments separately.** 

To start, we scoped the status checker to only show the first payment for PTR and ANCHOR. That's because we still needed to work out how to differentiate "regular" payments from "update" payments in the data — across all three programs, a resident could receive additional payments beyond what's normally expected (a single payment for ANCHOR/Senior Freeze, four for Stay NJ by design), and we didn't yet have a reliable way to identify, from the data itself, which payments were which. Now we do! [See full brief here](https://docs.google.com/document/d/15tj3fUNzD4sFdBIExjdPdMKLX8cntxRn1Um2tWTyGZo/edit?tab=t.0).

- According to [these designs](https://www.figma.com/design/fc59D8BV14x7EhW79ued5f/PAS-1-status-checker?node-id=1261-10901), we will show the regular “Direct deposit made” or “Check sent” content when a “regular payment” is made. 
- According to [these designs](https://www.figma.com/design/fc59D8BV14x7EhW79ued5f/PAS-1-status-checker?node-id=1261-10901), we will show “Your benefit amount was adjusted after further review, and a new check was sent on - **FINAL CONTENT** = "Your benefit amount was adjusted. A check was sent on [DATE]."

How to define "regular" v. "update" payments:

- For ANCHOR and Senior Freeze: the “regular payment” is the first payment made within each program. Any additional payments should be treated as “update payments.”

natekep:
## Link to ticket

This commit resolves [143](https://github.com/newjersey/tax-relief-status-checker/issues/143), but the deploy is on hold until team member is back from OOO. 

## LLM Disclaimer

## What was done?

- Across all three programs (ANCHOR/PTR), a resident could receive additional payments beyond what's normally expected, and we wanted to show to the resident the regular and updated payments, as defined in the notes section. 
- Created two helper functions, `showEarliestTransaction` and `showUpdatedTransaction` to return a JSX element table row based on whether a given transaction is a regular transaction (shows check or direct deposit details) or an update transaction (shows "your benefit amount was adjusted...").
- Created `showAllTransactions` helper function to find the earliest transaction and show it, followed by any update payments based on that program category. 
- Table row logic now uses `showAllTransactions` for each program category and quarter for Stay NJ transactions to show regular and update payments. 

## Accessibility

## Steps to test
 
- Created `testUtils.tsx` that contains quarterly transaction definitions for stay NJ payments, as well as other transaction definitions used for test cases in `payment-info/page.test.tsx`
- E2E tests with `npm run cypress`
- Unit tests with `npm run test`

- Before

Screenshot 2026-07-21 at 4 35 37 PM

- After

Screenshot 2026-07-21 at 4 36 59 PM

## Notes

How to define "regular" v. "update" payments:
For ANCHOR and Senior Freeze: the “regular payment” is the first payment made within each program. Any additional payments should be treated as “update payments.”
***Moved stay NJ into a separate ticket***


webapp/app/payment-info/page.test.tsx
```diff
@@ -1,15 +1,55 @@
...
+const checkTransactionInfo = (
+  rowHTML: string,
+  category: string,
+  date: string,
+  amount: number,
+  paymentType: PaymentType,
+) => {
+  if (paymentType === PaymentType.ADJUSTED) {
+    expect(rowHTML).toContain(`<td>${category}</td>`);
+    expect(rowHTML).toContain(`Your benefit amount was adjusted. A check was sent on ${date}`);
```
jachan, line 34: [DUST] Remind me again why we can't make a similar assertion here to what we assert on lines 46 and 50 with <td> tags?

webapp/app/payment-info/page.test.tsx
```diff
@@ -56,3 +96,119 @@ describe("getEarliestTransaction", () => {
     });
   });
 });
+
+describe("showEarliestTransaction", () => {
+  it("shows check sent on if method is check", () => {
```
jachan: [DUST] Consider "shows correct message when method is check"

webapp/app/payment-info/page.test.tsx
```diff
@@ -56,3 +96,119 @@ describe("getEarliestTransaction", () => {
...
+    checkTransactionInfo(
+      html,
+      "ANCHOR",
+      formatDate(payment_sent_transaction.payment_details.date),
+      payment_sent_transaction.payment_details.amount,
+      PaymentType.CHECK,
+    );
+  });
+
+  it("shows direct deposit made on if method is direct_deposit", () => {
```
jachan: [DUST] See comment on line 101

webapp/app/payment-info/page.test.tsx
```diff
@@ -56,3 +96,119 @@ describe("getEarliestTransaction", () => {
...
+    checkTransactionInfo(
+      html,
+      "ANCHOR",
+      formatDate(payment_sent_transaction.payment_details.date),
+      payment_sent_transaction.payment_details.amount,
+      PaymentType.DIRECT_DEPOSIT,
+    );
+  });
+
+  it("returns nothing if there is no payment_details", () => {
```
jachan, line 119: [DUST] nice test!

webapp/app/payment-info/page.tsx
```diff
@@ -28,6 +32,89 @@ export const getEarliestTransaction = (transactions: Transaction[]) => {
   return earliestTransaction;
 };
 
+export const showEarliestTransaction = (transaction: Transaction, category: string) => {
+  if (!transaction?.payment_details) return null;
+  return (
+    <tr>
+      <td>{category}</td>
+      {transaction.payment_details.method === "check" ? (
```
jachan: [DUST] do we have a type-safe guarantee that `transaction.payment_details.method` will always be either `check` or `direct_deposit`? This would be defined on the `Transaction`

If so, I'm okay with this ternary `else`, since our type guarantees that there are only two options. Otherwise, if `transaction.payment_details.method` is a string, we should change it into an enum.

webapp/app/payment-info/page.test.tsx
```diff
@@ -56,3 +96,119 @@ describe("getEarliestTransaction", () => {
...
+    expect(showEarliestTransaction(error_payment_sent_transaction, "ANCHOR")).toBeNull;
+  });
+});
+
+describe("showUpdatedTransaction", () => {
+  it("shows your benefit amount was adjusted if check has payment details", () => {
```
jachan: [DUST] see comments on 101/113 about naming, or an alternative is to use quotation marks i.e `"shows 'your benefit amount was adjusted' if check has payment details"`.

webapp/app/payment-info/page.tsx
```diff
@@ -28,6 +32,89 @@ export const getEarliestTransaction = (transactions: Transaction[]) => {
   return earliestTransaction;
 };
 
+export const showEarliestTransaction = (transaction: Transaction, category: string) => {
```
jachan: [PEBBLE] I can see a failure mode here where category is not one of [ANCHOR, PTR, Stay NJ] since right now it's a `string`. Suggest creating an enum for `TaxPrograms` and using it here for the category type.

webapp/app/payment-info/page.test.tsx
```diff
@@ -56,3 +96,119 @@ describe("getEarliestTransaction", () => {
...
+
+  it("returns nothing if there is no payment_details", () => {
+    expect(showUpdatedTransaction(error_payment_sent_transaction, "ANCHOR")).toBeNull;
+  });
+});
+
+describe("sortStayNJTransaction", () => {
+  describe("for ONE transaction", () => {
+    it("puts transaction with payment_date in bucket for range 1/1 to 4/31", () => {
+      expect(sortStayNJTransactions([stayQ1Start])).toEqual([[stayQ1Start], [], [], []]);
```
jachan: [PEBBLE] Suggest adding `expect(sortStayNJTransactions([stayQ1End])).toEqual([[stayQ1End], [], [], []]);` (and similar tests below) to test both ends of the Q1/Q2/etc range

webapp/app/payment-info/page.test.tsx
```diff
@@ -56,3 +96,119 @@ describe("getEarliestTransaction", () => {
...
+      PaymentType.DIRECT_DEPOSIT,
+    );
+  });
+
+  it("returns nothing if there is no payment_details", () => {
+    expect(showUpdatedTransaction(error_payment_sent_transaction, "ANCHOR")).toBeNull;
+  });
+});
+
+describe("sortStayNJTransaction", () => {
```
jachan: [DUST] Good job with the tests for this function, incredibly easy to read and understand

webapp/app/payment-info/page.tsx
jachan: [PEBBLE] This function also returns `JSX.Element | null `, right?

webapp/app/payment-info/page.tsx
```diff
@@ -28,6 +32,89 @@ export const getEarliestTransaction = (transactions: Transaction[]) => {
...
+
+export const showAllTransactions = (transactions: Transaction[], category: string) => {
+  const earliest = getEarliestTransaction(transactions);
+  if (!earliest?.payment_details) return null;
+
+  if (transactions.length === 1) {
+    return showEarliestTransaction(earliest, category);
+  }
+
+  const rest = transactions.filter((transaction) => transaction !== earliest);
```
jachan: [DUST] instead of `rest`, suggest `remainingTransactions`

webapp/app/payment-info/page.tsx
```diff
@@ -28,6 +32,89 @@ export const getEarliestTransaction = (transactions: Transaction[]) => {
...
+          Your benefit amount was adjusted. A check was sent on{" "}
+          {formatDate(transaction.payment_details.date)}
+        </div>
+      </td>
+      <td>${transaction.payment_details.amount}</td>
+    </tr>
+  );
+};
+
+export const showAllTransactions = (transactions: Transaction[], category: string) => {
```
jachan: [PEBBLE] I don't like `showAllTransactions` as a name here, because it's not showing ALL transactions, it's showing all transactions for a particular category. This method then needs to be called for every `supportedProgram`, and takes as input all the transactions in a particular Program. Important to name it as such so that future developers don't accidentally use it in the wrong way (e.g. passing the full list of transactions as input)

webapp/app/payment-info/page.tsx
jachan: In a similar vein, can we use the phrase `taxProgram` instead of `category` throughout this file? It's more descriptive and informative to a reader.

webapp/app/payment-info/page.tsx
```diff
@@ -28,6 +32,89 @@ export const getEarliestTransaction = (transactions: Transaction[]) => {
...
+
+  return (
+    <>
+      {showEarliestTransaction(earliest, category)}
+      {rest.map((transaction) => showUpdatedTransaction(transaction, category))}
+    </>
+  );
+};
+
+export const sortStayNJTransactions = (transactions: Transaction[]) => {
```
jachan: [DUST] This function, in combination with the comment on line 69, is making me think that it'd be useful to have a type for `Transaction[] + TaxProgram`. All functions would take that argument. The reason for this is to allow us to check what `TaxProgram` transactions belong to, and, in this case, exit early if `sortStayNJTransactions` is called for a set of transactions that ARE NOT StayNJ.

What do you think?

jachan: [PEBBLE] Let's add a return type here.

webapp/app/payment-info/page.tsx
```diff
@@ -103,33 +190,18 @@ const PaymentInfoPage = () => {
...
-                    <td>Senior Freeze</td>
-                    {transaction.payment_details.method === "check" ? (
-                      <td>Check issued on {formatDate(transaction.payment_details.date)}</td>
-                    ) : (
-                      <td>Direct deposit made on {formatDate(transaction.payment_details.date)}</td>
-                    )}
-                    <td>${transaction.payment_details.amount}</td>
-                  </tr>
-                );
+                if (ptr.length === 0) return null;
```
jachan, line 157: [DUST] Code looks great, appreciate you abstracting the logic into the helper functions.

webapp/app/payment-info/testUtils.tsx
```diff
@@ -0,0 +1,103 @@
...
+    method: "check",
+    check_number: "922775385",
+  },
+};
+
+export const stayQ1End = {
+  status: "payment_sent",
+  payment_details: {
+    amount: 377.56,
+    date: "04/30/2027 0:00:00",
```
jachan: [PEBBLE] Let's confirm this requirement with Emily before deploy - that time doesn't matter for these transactions, only the date. Also would love to confirm what timezone these dates are stored in (UTC vs local time) - that's a question for Juan and the data team.

webapp/app/payment-info/testUtils.tsx
```diff
@@ -0,0 +1,103 @@
...
+  payment_details: {
+    amount: 377.56,
+    date: "12/31/2027 0:00:00",
+    method: "check",
+    check_number: "922775385",
+  },
+};
+
+export const error_payment_sent_transaction = {
+  status: "payment_sent",
```
jachan: [PEBBLE] This seems like a weird test case - our API is incapable of producing a transaction like this, right?

webapp/app/page.tsx
```diff
@@ -37,6 +37,11 @@ const hasPaymentSentTransaction = (record: StatusRecord) => {
       return true;
     }
   }
+  for (const transaction of record.stay_nj) {
```
jachan: [DUST] One way to refactor this function would be to combine `record.anchor`, `record.ptr`, and `record.stay_nj` and then check to see if there are any transactions in that combined set with status `"payment_sent"`. Dries up the code slightly, but up to you if worth it or not.

webapp/cypress/e2e/paymentInfo.cy.ts
```diff
@@ -155,13 +156,44 @@ it("should route user to application-received if stay_nj", () => {
...
+
+    cy.intercept("POST", "/api/status", {
+      statusCode: 200,
+      body: resp,
+    });
+  });
+  cy.contains("button", `Check Status`).click();
+
+  cy.url().should("include", "/payment-info");
+  cy.contains("th", "Program").should("be.visible");
```
jachan: [DUST] Optional, but wondering if a helper method to assert certain conditions per row would be valuable in this test file - there's a lot of repeated code in each test.

jachan (commented):
I know we're not merging until Emily gets back, so no urgency on fixing these. Mostly requests for different code styling (some optional, some suggested) - logic itself looks solid and really readable.

webapp/app/payment-info/testUtils.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+export const earlier_transaction = {
+  status: "payment_sent",
+  payment_details: {
+    amount: 10.0,
+    date: "1/1/2026 0:00:00",
+    method: "direct_deposit",
+    check_number: "9NN775385",
+  },
+};
```
casewalker, line 22: dust - It feels odd having mixed camelCase and snake_case variables here, I would recommend sticking to only one or the other

webapp/app/page.tsx
casewalker: dust - Yeah, you could replace this function with:
```js
  return [...record.anchor, ...record.ptr, ...record.stay_nj].some(
    (transaction) => transaction["status"] === "payment_sent",
  );
```

webapp/app/page.tsx
```diff
@@ -37,6 +37,11 @@ const hasPaymentSentTransaction = (record: StatusRecord) => {
...
+    if (transaction["status"] === "payment_sent") {
```
casewalker: dust - Somewhat out of scope for this PR because it was the pre-existing state of the code, but the `anchor`, `ptr`, and `stay_nj` StatusRecord entries are currently just typed as `[]`, which leads the type of `transaction` here to be `never`, which means we're missing out on good type-checking.

webapp/app/payment-info/page.tsx
```diff
@@ -9,6 +9,10 @@ import { formatDate } from "../utils/formatDate";
 import { expandFaqAccordionItem, PaymentInfoPageFaq } from "@/components/PaymentInfoPageFaq";
 import { Transaction } from "@/components/types";
 
+const ptrString = "Senior Freeze";
+const anchorString = "ANCHOR";
+const stayNJString = "Stay NJ";
```
casewalker: sand - I think we tend to use all-caps (also known as screaming snake case) for global constants:
```suggestion
const PTR_STRING = "Senior Freeze";
const ANCHOR_STRING = "ANCHOR";
const STAYNJ_STRING = "Stay NJ";
```

webapp/app/payment-info/page.tsx
```diff
@@ -28,6 +32,89 @@ export const getEarliestTransaction = (transactions: Transaction[]) => {
...
+  );
+};
+
+export const showAllTransactions = (transactions: Transaction[], category: string) => {
+  const earliest = getEarliestTransaction(transactions);
+  if (!earliest?.payment_details) return null;
+
+  if (transactions.length === 1) {
+    return showEarliestTransaction(earliest, category);
+  }
```
casewalker: sand - This is an unnecessary/redundant guard clause. `rest` will correctly filter down to an empty array in the next line, and in the lower return statement the `map` won't execute it's body given empty input, so it will also return an empty array, which React doesn't render. So I would suggest removing this block entirely.

webapp/app/payment-info/page.tsx
```diff
@@ -28,6 +32,89 @@ export const getEarliestTransaction = (transactions: Transaction[]) => {
...
+  return (
+    <>
+      {showEarliestTransaction(earliest, category)}
+      {rest.map((transaction) => showUpdatedTransaction(transaction, category))}
+    </>
+  );
+};
+
+export const sortStayNJTransactions = (transactions: Transaction[]) => {
+  const sortedStayNJ: Transaction[][] = [[], [], [], []];
```
casewalker: dust/sand - I would rename this to call out the nested structure a bit, maybe something like `stayNJTransactionsByQuarter`?

webapp/app/payment-info/page.tsx
```diff
@@ -28,6 +32,89 @@ export const getEarliestTransaction = (transactions: Transaction[]) => {
...
+  for (const transaction of transactions) {
+    if (!transaction.payment_details) continue;
+    const currentDate = new Date(transaction.payment_details.date);
+    const currentMonth = currentDate.getMonth();
+    if (
+      currentMonth >= new Date("01/01").getMonth() &&
+      currentMonth <= new Date("04/30 23:59:59").getMonth()
+    ) {
```
casewalker: sand/pebble - I really don't know enough about this project to make any calls here, but this date-math feels brittle, and makes assumptions about the data.

1. As John called out separately, this may change behavior depending on the timezone of the machine running it and the timezone coming through on the transactions; that will be important to check. I think it would also be good to do hard-checks on the boundaries -- right now the tests have a 1-day granularity, but I would check either side of the boundary down to the seconds level. I tend to write dates in UTC at the Eastern boundary to be sure they are consistent across machines, like:
    ```js
    new Date("2026-05-01T03:59:59.000Z"); // 4/30 23:59:59 EDT
    // You can also just do the unix timestamp, e.g. `new Date(1777607999000)`
    // it is exact but generally more opaque
    ```

2. As well, this seems to implicitly assume the stay_nj data will always only be for the current year. Is that always true? On Jan 1st 2027, the data will only contain 2027 transactions? Maybe this is already true and confirmed and I just don't have the project context.

casewalker (commented):
Sorry it took me awhile to get to reviewing. Seems like good additions, a few comments here and there.

webapp/app/payment-info/page.tsx
casewalker: Shouldn't this be caught by a linter? Does this project really not have ESLint or Biome or anything? Shouldn't it?

natekep: Thanks for the callout we don't currently have a linter, but we have created a ticket for it [here](https://github.com/newjersey/tax-relief-status-checker/issues/186)

PR #202, opened 2026-08-07
[SHOW] - 160 added logging for a database row's DLN_NUM

### LINKED ISSUE #160: Start logging DLN_NUM on errors to help debug
# Acceptance Criteria
As an Innovation Engineer, whenever an error message occurs, I am able to see what DLN_NUM the error was associated with, allowing me to look up the exact record in the DB WITHOUT logging PII.

natekep:
## Link to ticket

This commit resolves [160](https://github.com/newjersey/tax-relief-status-checker/issues/160)

## LLM Disclaimer

- No LLM was used for this code

## What was done?

- Added logging to lambda function during the building of a record response. Done so that after a record row is accessed, an engineer is able to see what DLN_NUM the row was associated with, allowing them to look up the exact record in the DB without having to log PII. 

## Steps to test

- Log onto dev amplify and test API response with 020383155 | 08884  as ssn | zip pair then go to cloudwatch to confirm `DLN_NUM: 13800674901` is in the logs. Test with any other ssn/zip pairs you have from test db.

### DIFF
--- a/lambda/src/index.ts
+++ b/lambda/src/index.ts
@@ -96,6 +96,7 @@ const buildResponse = (rows: InquiryRow[]): BuildResponseResult => {
     return { records: [] };
   }
 
+  console.log(`DLN_NUM: ${rows[0].DLN_NUM}`);
   const records = rows.map(mapRowToRecord);
 
   return { records };
--- a/lambda/src/helpers.ts
+++ b/lambda/src/helpers.ts
@@ -1,5 +1,6 @@
 //Only used in tests
 export const buildMockRow = (overrides = {}) => ({
+  DLN_NUM: "00000001",
   SOCIAL_SECURITY_NUMBER_IDN: "123456789",
   ZIP_ADR: "12345",
   RETURN_YEAR_DTE: 2025,
--- a/lambda/src/types.d.ts
+++ b/lambda/src/types.d.ts
@@ -22,5 +22,6 @@ export interface InquiryRow {
   readonly RNY_APPLIED_DTE: string;
   readonly RETURN_YEAR_DTE: number;
   readonly TRANS_TOTAL_NUM: number;
+  readonly DLN_NUM: string;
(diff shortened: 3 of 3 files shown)


lambda/src/index.ts
```diff
@@ -96,6 +96,7 @@ const buildResponse = (rows: InquiryRow[]): BuildResponseResult => {
     return { records: [] };
   }
 
+  console.log(`DLN_NUM: ${rows[0].DLN_NUM}`);
```
jachan, line 99: DUST - What happens if this row isn't defined in the DB? Will this cause the entire API to fail, or just log an empty string/undefined? I think we want the latter behavior.

jachan (approved):
Such a simple change but should make a huge difference! Left a comment to make sure we don't accidentally take down the whole API.

lambda/src/index.ts
natekep: Will a record with a row that is not defined reach that log?

jachan: Unclear to me - I don't know what their DB schema is, and if that is required or not :(, that's why I'm trying to make sure our API can handle it

natekep: How can we test to make sure?

jachan: One way is to ask Emily to query the DB to see if there are any rows with a NULL value for DLN_NUM.

jachan: Another way is to see what the `rows[0]` object behaves if we query an attribute that does not exist. What happens if we ask to log `rows[0].NON_EXISTENT_ATTRIBUTE`? I would imagine that if a row didn't have a DLN_NUM, it would behave in the same way.

natekep: Screenshot 2026-08-07 at 4 04 55 PM

natekep: It gets logged as undefined instead of breaking anything

PR #213, opened 2026-08-10
[ASK] Add ANCHOR Autofile API

jachan:
## Link to ticket

#214 

## LLM Disclaimer

- Yes, an LLM generated the first draft of each piece of this PR: the infrastructure code, the lambda function, and the script to populate the DB.

## What was done?

- This PR creates an API that allows our website to determine if Taxation plans on auto-filing an ANCHOR application for a particular taxpayer.
  - This PR includes the CDK code to define the DynamoDB and the Lambda Function
  - This PR also includes the code for the lambda function to query the Dynamo DB
  - Additionally, it creates a script that allows engineers to populate the Dynamo DB table with csv data

- [NEW] After seeing the designs this morning, added the ability to tell if ANCHOR is going to be paid as Direct Deposit or a Check

- This feature is meant to be a scrappy, one-time data load from Excel Spreadsheets provided to us by Taxation, as we are working against a tight timeline (filers must notify Taxation by Sept 15th if they do not want auto-file to happen)

## Accessibility
N/A, no visual changes

## Steps to test
- Deployed to dev and tested in dev:
  - Created test data that matches prod schema
  - ran `npm run deploy:dev`
  - ran helper script `npx populate-db-from-csv.ts` to load data into DB
  - tested lambda from console to validate that autofile API returns true for the test data record
  - tested lambda from console to validate that autofile API returns false for a record not in the test data

image
image

image

<details>
<summary>Testing the API with a record in the test data with Direct Deposit data in the CSV</summary>
image
image
</details>

<details>
<summary>Testing the API with a record in the test data without Direct Deposit data in the CSV</summary>
image
image

</details>

<details>
<summary>Testing the API with a record NOT in the test data</summary>
image
image

</details>

## Notes
After this API is implemented, the next PR will be to update the website to call this API and then display conditional content based on if the taxpayer's ANCHOR autofile status (pending designs)

### DIFF
--- a/lambda-anchor-autofile-api/scripts/hashSsnZip.ts
+++ b/lambda-anchor-autofile-api/scripts/hashSsnZip.ts
@@ -5,7 +5,7 @@
  *
  * Example: npx tsx scripts/hash-ssn-zip.ts 123456789 07001
  */
-import { computeSsnZipHash } from "../src/index.js";
+import { computeSsnZipHash } from "../src/util/computeSsnZipHash";
 
 const SSN_PATTERN = /^\d{9}$/;
 const ZIP_PATTERN = /^\d{5}$/;
--- a/lambda-anchor-autofile-api/scripts/populate-db-from-csv.ts
+++ b/lambda-anchor-autofile-api/scripts/populate-db-from-csv.ts
@@ -1,0 +1,191 @@
+/**
+ * Reads a CSV file containing TPID and Zip columns, extracts SSN (characters 1–9 of TPID) and ZIP
+ * (first 5 characters of Zip column), hashes them together, and writes the resulting ssnZipHash
+ * values to the DynamoDB autofile table.
+ *
+ * Usage: npx tsx scripts/seed-from-csv.ts <path-to-csv> [--table <table-name>] [--region <region>]
+ *
+ * Example: npx tsx scripts/seed-from-csv.ts ~/Desktop/autofile-test-data.csv --table
+ * anchor-autofile-ssn-zip-dev
+ */
+import { createReadStream } from "node:fs";
+import { createInterface } from "node:readline";
+import { DynamoDBClient, BatchWriteItemCommand } from "@aws-sdk/client-dynamodb";
+import type { WriteRequest } from "@aws-sdk/client-dynamodb";
+import { computeSsnZipHash } from "../src/index.js";
+
+/** Configuration parsed from CLI arguments. */
+interface SeedConfig {
+  /** Path to the input CSV file. */
+  readonly csvPath: string;
+
+  /** DynamoDB table name. */
+  readonly tableName: string;
+
+  /** AWS region for the DynamoDB client. */
+  readonly region: string;
+}
+
+/** Maximum items per DynamoDB BatchWriteItem call. */
+const BATCH_SIZE = 25;
+
+const SSN_PATTERN = /^\d{9}$/;
+const ZIP_PATTERN = /^\d{5}$/;
+
+/** Parses CLI arguments into a SeedConfig. */
+const parseArgs = (): SeedConfig => {
+  const args = process.argv.slice(2);
+  const csvPath = args.find((arg) => !arg.startsWith("--"));
+
+  if (!csvPath) {
+    console.error(
+      "Usage: npx tsx scripts/seed-from-csv.ts <path-to-csv> [--table <name>] [--region <region>]",
+    );
+    process.exit(1);
+  }
+
+  const tableIndex = args.indexOf("--table");
+  const tableName = tableIndex !== -1 ? args[tableIndex + 1] : "anchor-autofile-ssn-zip-dev";
+
+  const regionIndex = args.indexOf("--region");
+  const region = regionIndex !== -1 ? args[regionIndex + 1] : "us-east-1";
+
+  return { csvPath, tableName, region };
+};
+
+/** Extracts SSN from a TPID value (characters at index 1–9). */
+const extractSsnFromTpid = (tpid: string): string => {
+  return tpid.slice(1, 10);
+};
+
+/** Extracts 5-digit ZIP from the Zip column value. */
+const extractZipCode = (rawZip: string): string => {
+  return rawZip.slice(0, 5);
+};
+
+/** Sends a batch of write requests to DynamoDB. */
+const writeBatch = async (
+  client: DynamoDBClient,
+  tableName: string,
+  requests: WriteRequest[],
+): Promise<void> => {
+  const command = new BatchWriteItemCommand({
+    RequestItems: { [tableName]: requests },
+  });
+
+  const result = await client.send(command);
+  const unprocessed = result.UnprocessedItems?.[tableName];
+
+  if (unprocessed && unprocessed.length > 0) {
+    console.warn(`Retrying ${unprocessed.length} unprocessed items from batch...`);
+    await writeBatch(client, tableName, unprocessed);
+  }
+};
+
+/** Reads the CSV and returns hashed write requests. */
+const processFile = async (csvPath: string): Promise<WriteRequest[]> => {
+  const stream = createReadStream(csvPath, { encoding: "utf-8" });
+  const lines = createInterface({ input: stream });
+
+  const requests: WriteRequest[] = [];
+  let headerParsed = false;
+  let tpidIndex = -1;
+  let zipIndex = -1;
+  let lineNumber = 0;
+  let skipped = 0;
+
+  for await (const line of lines) {
+    lineNumber++;
+
+    const columns = line.replace(/^\uFEFF/, "").split(",");
+
+    if (!headerParsed) {
+      tpidIndex = columns.findIndex((col) => col.trim().toUpperCase() === "TPID");
+      zipIndex = columns.findIndex((col) => col.trim().toUpperCase() === "ZIP");
+
+      if (tpidIndex === -1 || zipIndex === -1) {
+        console.error(
+          `CSV must contain TPID and Zip columns. Found headers: ${columns.join(", ")}`,
+        );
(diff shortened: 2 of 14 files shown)


lambda-anchor-autofile-api/scripts/hashSsnZip.ts
jachan, line 1: This script creates a hash from SSN + ZIP, so I could manually put it in the DB and test the API. Curious if there's still value in keeping it around. I could see it being useful in the future to be able to debug (put in an SSN+ZIP to get the hash to query/check the DB manually)

package.json
```diff
@@ -9,7 +9,8 @@
     "dev": "cd webapp && next dev --port 3000",
     "build": "cd webapp && next build",
     "cypress": "cd webapp && cypress run",
-    "package": "cd lambda && npm run package",
+    "package": "cd lambda && npm run package && cd ../lambda-anchor-autofile-api && npm run package",
+    "package:autofile": "cd lambda-anchor-autofile-api && npm run package",
```
aloverso: You can do this using `prefix` instead which I would recommend, I think having to introduce `cd ..` is somewhat brittle

`package: "npm run package --prefix lambda && npm run package --prefix lambda-anchor-autofile-api"`

lambda-anchor-autofile-api/src/index.ts
```diff
@@ -0,0 +1,72 @@
...
+};
+
+/** Validates that the event contains a valid SSN and ZIP. */
+const validateEvent = (event: unknown): AutofileLookupRequest | null => {
+  if (!event || typeof event !== "object") return null;
+
+  const { ssn, zip } = event as Record<string, unknown>;
+
+  if (typeof ssn !== "string" || !/^\d{9}$/.test(ssn)) return null;
+  if (typeof zip !== "string" || !/^\d{5}$/.test(zip)) return null;
```
aloverso, line 29: [dust] Does this really matter? If someone inputs a zip with the wrong number of digits, then the hash will just be different and the lookup will return no results found. Seems like it doesn't provide much value to check it up front, especially since it's just getting called by direct invocation.

lambda-anchor-autofile-api/src/index.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  it("returns autofilePlanned false when item does not exist", async () => {
+    dynamoMock.on(GetItemCommand).resolves({
+      Item: undefined,
+    });
+
+    const result = await handler({ ssn: "123456789", zip: "07001" });
+
+    expect(result).toEqual({ autofilePlanned: false });
+  });
```
aloverso: [sand] As a refactor in the future, I think it's generally a good idea to use some kind of fake dynamo rather than a mock, which actually tests the validity of the Get and Insert statements.

lambda-anchor-autofile-api/scripts/populate-db-from-csv.ts
```diff
@@ -0,0 +1,191 @@
...
+interface SeedConfig {
+  /** Path to the input CSV file. */
+  readonly csvPath: string;
+
+  /** DynamoDB table name. */
+  readonly tableName: string;
+
+  /** AWS region for the DynamoDB client. */
+  readonly region: string;
+}
```
aloverso: [dust] You know how I feel about [code comments](https://scaling-adventure-v9318vm.pages.github.io/blog/code-comments/) so feel free to take my opinion as you will. And I know from the disclaimer this is LLM-generated. 

But I personally find this very unreadable, to have to look at all these extra comments when skimming a file. I would find this difficult to work in and maintain. We as developers should seek to reduce complexity and increase clarity in our code. And I would argue that "Path to the input CSV file." as a label for `csvPath` is redundant noise that increases complexity. 

If "AWS region for the DynamoDB client" is truly important context for `region` then a good practice would be to name it `awsDynamoRegion` rather than adding the cognitive load of reading (and maintaining) a code comment

lambda-anchor-autofile-api/scripts/populate-db-from-csv.ts
```diff
@@ -0,0 +1,191 @@
...
+  const regionIndex = args.indexOf("--region");
+  const region = regionIndex !== -1 ? args[regionIndex + 1] : "us-east-1";
+
+  return { csvPath, tableName, region };
+};
+
+/** Extracts SSN from a TPID value (characters at index 1–9). */
+const extractSsnFromTpid = (tpid: string): string => {
+  return tpid.slice(1, 10);
+};
```
aloverso: [dust] The one thing that I feel could actually benefit from the clarity of a code comment to explain - I don't know what is a TPID / what does it stand for

lambda-anchor-autofile-api/scripts/populate-db-from-csv.ts
```diff
@@ -0,0 +1,191 @@
...
+    );
+    process.exit(1);
+  }
+
+  const tableIndex = args.indexOf("--table");
+  const tableName = tableIndex !== -1 ? args[tableIndex + 1] : "anchor-autofile-ssn-zip-dev";
+
+  const regionIndex = args.indexOf("--region");
+  const region = regionIndex !== -1 ? args[regionIndex + 1] : "us-east-1";
+
```
aloverso: [dust] Would recommend putting these defaults into global variables here.

Maybe there's also something you could do with the command printout to let the user know what the defaults are when it is being used.

lambda-anchor-autofile-api/scripts/populateDbFromCsv.ts
```diff
@@ -0,0 +1,191 @@
...
+    RequestItems: { [tableName]: requests },
+  });
+
+  const result = await client.send(command);
+  const unprocessed = result.UnprocessedItems?.[tableName];
+
+  if (unprocessed && unprocessed.length > 0) {
+    console.warn(`Retrying ${unprocessed.length} unprocessed items from batch...`);
+    await writeBatch(client, tableName, unprocessed);
+  }
```
aloverso, line 77: [dust] Infinite loop? I guess a dev would just ctrl-C if they saw this getting printed forever?

lambda-anchor-autofile-api/scripts/populate-db-from-csv.ts
```diff
@@ -0,0 +1,191 @@
...
+      continue;
+    }
+
+    const tpid = columns[tpidIndex]?.trim();
+    const rawZip = columns[zipIndex]?.trim();
+
+    if (!tpid || !rawZip) {
+      skipped++;
+      continue;
+    }
```
aloverso: [dust/pebble] Why not use a library here like [csv-parse](https://www.npmjs.com/package/csv-parse). I think there's a lot of room for error in rolling your own parser to account for BOM, for example, and this is a solved problem by these libaries

lambda-anchor-autofile-api/scripts/populate-db-from-csv.ts
```diff
@@ -0,0 +1,191 @@
...
+ * Usage: npx tsx scripts/seed-from-csv.ts <path-to-csv> [--table <table-name>] [--region <region>]
+ *
+ * Example: npx tsx scripts/seed-from-csv.ts ~/Desktop/autofile-test-data.csv --table
+ * anchor-autofile-ssn-zip-dev
+ */
+import { createReadStream } from "node:fs";
+import { createInterface } from "node:readline";
+import { DynamoDBClient, BatchWriteItemCommand } from "@aws-sdk/client-dynamodb";
+import type { WriteRequest } from "@aws-sdk/client-dynamodb";
+import { computeSsnZipHash } from "../src/index.js";
```
aloverso: [pebble] Feels weird to directly import this function from a lambda handler file. Could you extract it into a shared util that both this script and the lambda can import from?

jachan:
Thanks @aloverso - originally prioritized speed here to get something merged (ANCHOR mailers went out TODAY and filers will start wondering what's going on!) 

Your comments made me realize that I probably should have been more thoughtful about revising the LLM generated code, moved this PR into draft and will work on it tomorrow morning.

package.json
jachan: Wasn't aware of this, thanks for the suggestion!

lambda-anchor-autofile-api/scripts/populate-db-from-csv.ts
jachan: I usually go through and remove extraneous LLM comments, I skipped it for this because of the one-off nature of this script. Agree that they are generally not helpful for human review.

lambda-anchor-autofile-api/scripts/populate-db-from-csv.ts
jachan: "Taxpayer ID", I will add a comment

lambda-anchor-autofile-api/scripts/populate-db-from-csv.ts
jachan: Have always been unsure when to use a library vs using the default JS libraries to do the same thing. But yes, agree that this is probably a good usecase for a library, thanks for the suggestion!

lambda-anchor-autofile-api/src/index.ts
jachan: This is getting invoked by a call through a Lambda Function URL, rather than being invoked through the Lambda API, so I think some level of validation is still worthwhile.

jachan:
Learned this morning in design review that we needed to add functionality to the API to determine whether a ANCHOR Autofile application is going to be sent as a CHECK or DIRECT_DEPOSIT. Updated the logic and helper script accordingly.

lambda-anchor-autofile-api/src/index.test.ts
jachan: Open to doing this in a followup, but it does seem like a lot of work to set up!

PR #225, opened 2026-08-13
[SHOW] - Created Logout button component and moved into header

### LINKED ISSUE #172: Revisiting the CTA "Log Out" copy + placement

natekep:
## Link to ticket

This commit resolves [172](https://github.com/newjersey/tax-relief-status-checker/issues/172)

## LLM Disclaimer

- Used an LLM to figure out how to access pathname

## What was done?

- Created a `LogoutButton.tsx` component to remove logout button from each page besides landing and into layout
- `LogoutButton` renders conditionally based off pathname (only want it on /more-information-needed", "/application-received", "/payment-info)
- Moved `LogoutButton.tsx` component into `TaxStatusCheckerHeader.tsx` component to match its location in mocks

## Accessibility

- [ ] I have made sure this works with a screenreader!
- [ ] I have checked this with the [WAVE extension](https://wave.webaim.org/extension/).

## Steps to test

- Manually test

- Before
Screenshot 2026-08-13 at 4 24 55 PM

- After
Screenshot 2026-08-13 at 4 26 41 PM

## Notes

Before merge - see if there is a way to remove raw styling `textAlign: right`: see https://github.com/newjersey/tax-relief-status-checker/pull/236#discussion_r3807816380

### DIFF
--- a/webapp/components/LogoutButton.tsx
+++ b/webapp/components/LogoutButton.tsx
@@ -1,0 +1,26 @@
+"use client";
+import { usePathname } from "next/navigation";
+import Link from "next/link";
+
+const PAGES_WITH_LOGOUT = ["/more-information-needed", "/application-received", "/payment-info"];
+
+export const LogoutButton = () => {
+  const pathname = usePathname();
+
+  const shouldShowButton = PAGES_WITH_LOGOUT.includes(pathname);
+
+  if (!shouldShowButton) return null;
+
+  return (
+    <>
+      <div style={{ textAlign: "right" }}>
+        <Link className="usa-button usa-button--outline margin-right-3 margin-top-3" href="/">
+          <svg focusable="false" role="img" width="20" height="20" fill="#005ea2">
+            <use href="/img/sprite.svg#logout"></use>
+          </svg>
+          Log out
+        </Link>
+      </div>
+    </>
+  );
+};
--- a/webapp/components/LogoutButton.tsx
+++ b/webapp/components/LogoutButton.tsx
@@ -13,13 +13,15 @@ export const LogoutButton = () => {
 
   return (
     <>
-      <div style={{ textAlign: "right" }}>
-        <Link className="usa-button usa-button--outline margin-right-3 margin-top-3" href="/">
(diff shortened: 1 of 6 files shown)


jachan (approved):
Thanks for getting this in!

webapp/components/LogoutButton.tsx
```diff
@@ -0,0 +1,28 @@
+"use client";
+import { usePathname } from "next/navigation";
+import Link from "next/link";
+
+const PAGES_WITH_LOGOUT = ["/more-information-needed", "/application-received", "/payment-info"];
```
jachan: [PEBBLE] Can we flip this? So instead of defining PAGES_WITH_LOGOUT we define pages WITHOUT logout? That way when we add new pages they will by default have the logout button.

jachan: and with that change, i'm curious what would happen if we tried to navigate to a page that did not exist - for example: https://propertytaxreliefstatus.nj.gov/not-found

natekep: Screenshot 2026-08-14 at 11 22 03 AM
honestly makes it a little bit better as the header still gets rendered on every page and it gives them a way to get back to landing, I think the "log out" is a little confusing language however

jachan: Yeah, right now, since we don't have a `not-found` file, we're rendering the default NextJS not found page, and then the Header gets put on top of that. I do think having "Log out" on the page might be really confusing to users though.

Curious if it'd be better to redirect any 404 to the index page, or if that's an a11y or web bad practice.

natekep: Is that something that is out of scope for this PR? Im imagining we have to get the mocks for that `not-found` page, but once that is done just add that pathname to the `PAGES_WITHOUT_LOGOUT` list?

natekep: Or go back to `PAGES_WITH_LOGOUT` for now until the `not-found` page is made?

jachan:
Closed in favor of rebased PR here: https://github.com/newjersey/tax-relief-status-checker/pull/255

PR #236, opened 2026-08-17
214 - Add autofile anchor webpages

### LINKED ISSUE #214: ANCHOR Autofile Notice
**Goal:** Provide a self-service, web-based way for NJ residents to find out whether the State is planning to autofile an ANCHOR application for them (based on a one-time data dump provided by the Division of Taxation). [Full brief here](https://docs.google.com/document/d/15DYTcrxLkvxiUoU_mTybsFb_zeCbLWq4v0iZr-o68m8/edit?tab=t.0).

**Requirements**
Check data dump first: if an entered SSN & ZIP pair match a pair found in the data dump, return auto-file notice.
* If resident received a check last year, show [THIS DESIGN](https://www.figma.com/design/fc59D8BV14x7EhW79ued5f/PAS-1-status-checker?node-id=1782-4554&t=HIK0eqeXMG7P24SY-4) ("payment type" = check and specific FAQs).
* If a resident used direct deposit last year, show [THIS DESIGN](https://www.figma.com/design/fc59D8BV14x7EhW79ued5f/PAS-1-status-checker?node-id=1782-4590&t=HIK0eqeXMG7P24SY-4) ("payment type" = direct deposit and specific FAQs).

Otherwise (if an entered SSN and zip pair don't match a pair found in the data dump), behavior should remain the same:
1. If no match found in elf_saver_inquiry, show "no record found" alert
2. If match found but no payment made yet, show current "processing page," 
3. If match found and payment made, show "payment info" page.

**Acceptance Criteria**
* Enter SSN & ZIP from data dump that has banking info populated and confirm that autofile design shows with "Payment Type" = Direct Deposit and the following FAQs show: 1) "What if I need to change my banking information?" and 2) "How can I switch from direct deposit payment to a physical check?"
* Enter SSN & ZIP from data dump that has banking info populated and confirm that autofile design shows with "Payment Type" = Check and the following FAQs show: 1) "What if I need to change my mailing address?" and 2) "What if I want to switch from a physical check to a direct deposit payment?"
* Enter SSN & ZIP from elf_saver_inquiry that has a Senior Freeze transaction with status = payment_sent and confirm that "payment info" page shows.
* Enter SSN & ZIP from elf_saver_inquiry that has a Senior Freeze transactions with status = processing and confirm that "processing" page shows.
* Enter SSN & ZIP not found in elf_saver_inquiry or data dump and confirm that "No 2025 application found" alert shows.

jachan:
## Link to ticket

This commit resolves #214 

## LLM Disclaimer

- Yes, an LLM was used to debug and fix the SSN validation code in `lambda-anchor-autofile-api`

## What was done?
Mailers have gone out to 1.1M+ taxpayers who are a part of the ANCHOR Autofile population. Taxation has provided us with a list of the taxpayers, and we have populated a static DynamoDB DB with hashed SSN+Zip pairs in PR #213.

This PR adds the `api/autofile` route to the website, so we can proxy the request to our lambda function. It also adds the `/anchor-autofile` route, which is shown to taxpayers who are found in our DynamoDB table. One important feature is that based on whether their planned `paymentType` is `check` or `direct_deposit`, the FAQs for the `anchor-autofile` page change accordingly

## Accessibility

## Steps to test
- A reviewer can test using non-prod data by going to the following URL: https://214-add-autofile-anchor-webpages.d1fv5gc7ba4m6a.amplifyapp.com/

The following non-prod SSN/Zip pairs can be used to test the new functionality:

Check
```
{
  "ssn": "123456789",
  "zip": "11111"
}
```

Direct Deposit
```
{
  "ssn": "123456789",
  "zip": "12345"
}
```

image

image

## Notes
The plan is to remove this feature on Sept 15th, as after that point, taxpayers can no longer correct their autofile status. After this feature is removed, we will return to just using the Status Lookup Lambda.

### DIFF
--- a/infra/lib/infra-stack.ts
+++ b/infra/lib/infra-stack.ts
@@ -91,21 +91,6 @@ export class InfraStack extends Stack {
       assumedBy: new iam.ServicePrincipal("amplify.amazonaws.com"),
     });
 
-    amplifyIamRole.attachInlinePolicy(
-      new iam.Policy(this, "InvokeLambdaPolicy", {
-        statements: [
-          new iam.PolicyStatement({
-            actions: ["lambda:InvokeFunctionUrl", "lambda:InvokeFunction"],
-            resources: [this.lambdaFunction.functionArn],
-          }),
-          new iam.PolicyStatement({
-            actions: ["lambda:InvokeFunctionUrl", "lambda:InvokeFunction"],
-            resources: [this.autofileLambdaFunction.functionArn],
-          }),
-        ],
-      }),
-    );
-
     const customPolicy = new iam.Policy(this, "LambdaCustomPolicy", {
       statements: [
         new iam.PolicyStatement({
@@ -145,6 +130,17 @@ export class InfraStack extends Stack {
       authType: lambda.FunctionUrlAuthType.AWS_IAM,
       cors: { allowedOrigins: props.allowedOrigins },
     });
+
+    amplifyIamRole.attachInlinePolicy(
+      new iam.Policy(this, "InvokeLambdaPolicy", {
+        statements: [
+          new iam.PolicyStatement({
+            actions: ["lambda:InvokeFunctionUrl", "lambda:InvokeFunction"],
+            resources: [this.lambdaFunction.functionArn, this.autofileLambdaFunction.functionArn],
+          }),
+        ],
+      }),
+    );
   }
 
   private buildVpcConfig(
--- a/infra/lib/infra-stack.ts
+++ b/infra/lib/infra-stack.ts
@@ -98,6 +98,10 @@ export class InfraStack extends Stack {
             actions: ["lambda:InvokeFunctionUrl", "lambda:InvokeFunction"],
             resources: [this.lambdaFunction.functionArn],
           }),
+          new iam.PolicyStatement({
+            actions: ["lambda:InvokeFunctionUrl", "lambda:InvokeFunction"],
+            resources: [this.autofileLambdaFunction.functionArn],
+          }),
         ],
       }),
     );
--- a/lambda-anchor-autofile-api/src/index.test.ts
+++ b/lambda-anchor-autofile-api/src/index.test.ts
@@ -35,6 +35,17 @@ describe("handler", () => {
       expect(result).toEqual({ error: "Request must include ssn (9 digits) and zip (5 digits)" });
     });
 
+    it("strips hyphens from SSN before validation", async () => {
+      dynamoMock.on(GetItemCommand).resolves({ Item: undefined });
+
+      const result = await handler({ ssn: "123-45-6789", zip: "07001" });
+
+      expect(result).toEqual({ autofilePlanned: false });
+      const expectedHash = computeSsnZipHash("123456789", "07001");
+      const call = dynamoMock.commandCalls(GetItemCommand)[0];
+      expect(call.args[0].input.Key).toEqual({ ssnZipHash: { S: expectedHash } });
+    });
+
     it("returns error for null event", async () => {
       const result = await handler(null);
 
--- a/lambda-anchor-autofile-api/src/index.test.ts
+++ b/lambda-anchor-autofile-api/src/index.test.ts
@@ -35,23 +35,6 @@ describe("handler", () => {
       expect(result).toEqual({ error: "Request must include ssn (9 digits) and zip (5 digits)" });
     });
 
-    it("strips hyphens from SSN before validation", async () => {
-      dynamoMock.on(GetItemCommand).resolves({ Item: undefined });
-
-      const result = await handler({ ssn: "123-45-6789", zip: "07001" });
-
-      expect(result).toEqual({ autofilePlanned: false });
-      const expectedHash = computeSsnZipHash("123456789", "07001");
-      const call = dynamoMock.commandCalls(GetItemCommand)[0];
-      expect(call.args[0].input.Key).toEqual({ ssnZipHash: { S: expectedHash } });
-    });
-
-    it("returns error for null event", async () => {
-      const result = await handler(null);
-
-      expect(result).toEqual({ error: "Request must include ssn (9 digits) and zip (5 digits)" });
-    });
-
     it("throws when TABLE_NAME is not configured", async () => {
       vi.stubEnv("TABLE_NAME", "");
 
@@ -69,7 +52,7 @@ describe("handler", () => {
 
       const result = await handler({ ssn: "123456789", zip: "07001" });
 
-      expect(result).toEqual({ autofilePlanned: true, paymentMethod: "CHECK" });
+      expect(result).toEqual({ autofilePlanned: true, paymentMethod: "check" });
     });
 
     it("returns autofilePlanned true with DIRECT_DEPOSIT when item has direct deposit", async () => {
@@ -79,7 +62,7 @@ describe("handler", () => {
 
       const result = await handler({ ssn: "123456789", zip: "07001" });
 
-      expect(result).toEqual({ autofilePlanned: true, paymentMethod: "DIRECT_DEPOSIT" });
+      expect(result).toEqual({ autofilePlanned: true, paymentMethod: "direct_deposit" });
     });
 
     it("returns autofilePlanned false when item does not exist", async () => {
--- a/lambda-anchor-autofile-api/src/index.ts
+++ b/lambda-anchor-autofile-api/src/index.ts
@@ -1,9 +1,9 @@
 import { DynamoDBClient, GetItemCommand } from "@aws-sdk/client-dynamodb";
 import { computeSsnZipHash } from "./util/computeSsnZipHash";
 
-enum PaymentMenthod {
-  Check = "CHECK",
-  DirectDeposit = "DIRECT_DEPOSIT",
+enum PaymentMethod {
+  Check = "check",
+  DirectDeposit = "direct_deposit",
 }
 
 export interface AutofileLookupRequest {
@@ -14,7 +14,7 @@ export interface AutofileLookupRequest {
 export interface AutofileLookupResponse {
   /** Whether the SSN/ZIP pair is planned for ANCHOR autofile. */
   readonly autofilePlanned: boolean;
-  readonly paymentMethod?: PaymentMenthod;
+  readonly paymentMethod?: PaymentMethod;
 }
 
 const dynamoClient = new DynamoDBClient({});
@@ -69,6 +69,6 @@ export const handler = async (
 
   return {
     autofilePlanned: true,
-    paymentMethod: isDirectDeposit ? PaymentMenthod.DirectDeposit : PaymentMenthod.Check,
+    paymentMethod: isDirectDeposit ? PaymentMethod.DirectDeposit : PaymentMethod.Check,
   };
 };
--- a/lambda-anchor-autofile-api/src/index.ts
+++ b/lambda-anchor-autofile-api/src/index.ts
@@ -25,10 +25,13 @@ const validateEvent = (event: unknown): AutofileLookupRequest | null => {
 
   const { ssn, zip } = event as Record<string, unknown>;
 
-  if (typeof ssn !== "string" || !/^\d{9}$/.test(ssn)) return null;
-  if (typeof zip !== "string" || !/^\d{5}$/.test(zip)) return null;
+  if (typeof ssn !== "string" || typeof zip !== "string") return null;
 
-  return { ssn, zip };
+  const sanitizedSsn = ssn.replace(/-/g, "");
+  if (!/^\d{9}$/.test(sanitizedSsn)) return null;
+  if (!/^\d{5}$/.test(zip)) return null;
+
+  return { ssn: sanitizedSsn, zip };
 };
 
 /** Response shape returned when validation fails. */
--- a/lambda-anchor-autofile-api/src/index.ts
+++ b/lambda-anchor-autofile-api/src/index.ts
@@ -19,11 +19,9 @@ export interface AutofileLookupResponse {
 
 const dynamoClient = new DynamoDBClient({});
 
-/** Validates that the event contains a valid SSN and ZIP. */
-const validateEvent = (event: unknown): AutofileLookupRequest | null => {
-  if (!event || typeof event !== "object") return null;
-
-  const { ssn, zip } = event as Record<string, unknown>;
+const validateEvent = (event: Record<string, unknown>): AutofileLookupRequest | null => {
+  const body = typeof event.body === "string" ? JSON.parse(event.body) : event;
+  const { ssn, zip } = body;
 
   if (typeof ssn !== "string" || typeof zip !== "string") return null;
 
--- a/webapp/app/anchor-autofile/AnchorAutofileCheckFaqContent.tsx
+++ b/webapp/app/anchor-autofile/AnchorAutofileCheckFaqContent.tsx
@@ -6,20 +6,9 @@ export const AnchorAutofileCheckFaqContent: FaqItem[] = [
     content: (
       <>
         <p>
-          That's okay, nothing's wrong. The State is autofiling on your behalf regardless of whether
-          you received a letter. If you need to change your payment method or mailing address,
-          you'll need to file an application manually{" "}
-          <a href="https://propertytaxreliefapp.nj.gov/">online</a> or by mailing{" "}
-          <a href="https://www.nj.gov/treasury/taxation/propertyreliefforms.shtml">
-            a paper application
-          </a>{" "}
-          to:{" "}
+          In your case, the Division is filing on your behalf. You do not need to file an ANCHOR
+          application.
         </p>
-        <p>NJ Division of Taxation</p>
-        <p>Revenue Processing Center Property</p>
-        <p>Tax Relief Application</p>
-        <p>PO Box 636</p>
-        <p>Trenton, NJ 08646-0636</p>
       </>
     ),
     expanded: false,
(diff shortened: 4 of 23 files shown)


lambda-anchor-autofile-api/src/index.ts
```diff
@@ -1,9 +1,9 @@
 import { DynamoDBClient, GetItemCommand } from "@aws-sdk/client-dynamodb";
 import { computeSsnZipHash } from "./util/computeSsnZipHash";
 
-enum PaymentMenthod {
```
jachan, line 4: Fixed typo!

webapp/cypress/e2e/landingPage.cy.ts
```diff
@@ -6,6 +6,10 @@ beforeEach(() => {
     win.gtag = cy.stub().as("gtag");
   });
   cy.visit("/");
+  cy.intercept("POST", "/api/autofile", {
```
jachan, line 9: This intercept (and the ones below it) are needed because we only call the status lookup lambda if the autofile API returns "not_planned"

webapp/components/TaxReliefDataProvider.tsx
```diff
@@ -2,9 +2,22 @@
...
+import { IssueFlaggedType, PaymentMethod, Transaction } from "./types";
 
-export interface TaxReliefStatusData {
+export enum DataType {
+  AUTOFILE,
+  STATUS,
+}
+
+export interface AutofileData {
+  readonly kind: DataType.AUTOFILE;
```
aloverso: [dust] Maybe it's just me, but `kind` feels like an odd choice of term here. I'd prefer `type`

webapp/app/api/autofile/route.ts
```diff
@@ -0,0 +1,85 @@
...
+};
+
+/**
+ * POST /api/status
+ *
+ * Proxies the SSN + ZIP lookup to the AWS Lambda endpoint configured via
+ * NEXT_PUBLIC_AUTOFILE_LAMBDA_API_URL. Signs the request with SigV4 for AWS_IAM authentication on
+ * the Lambda Function URL.
+ */
+export const POST = async (request: NextRequest): Promise<NextResponse> => {
```
aloverso, line 48: [dust] Does this file have a test?

aloverso (commented):
Looks good. Do you have tests for most of the pages and components shown here? I would have expected some changes to the tests based on the contents of the pages changing.

lambda-anchor-autofile-api/src/index.ts
casewalker: my favorite type of cigarette, menthods.

infra/lib/infra-stack.ts
```diff
@@ -98,6 +98,10 @@ export class InfraStack extends Stack {
             actions: ["lambda:InvokeFunctionUrl", "lambda:InvokeFunction"],
             resources: [this.lambdaFunction.functionArn],
           }),
+          new iam.PolicyStatement({
+            actions: ["lambda:InvokeFunctionUrl", "lambda:InvokeFunction"],
+            resources: [this.autofileLambdaFunction.functionArn],
+          }),
```
casewalker: sand - I don't see any reason not to combine two statements into one:

```js
          new iam.PolicyStatement({
            actions: ["lambda:InvokeFunctionUrl", "lambda:InvokeFunction"],
            resources: [this.lambdaFunction.functionArn, this.autofileLambdaFunction.functionArn],
          }),
```

webapp/components/TaxReliefDataProvider.tsx
```diff
@@ -2,9 +2,22 @@
...
+  readonly lastFourSsnDigits: string;
+  readonly zipCode: string;
+  readonly paymentMethod?: PaymentMethod;
+}
+
+export interface StatusData {
```
casewalker, line 19: dust - Probably not important to do, but these two types are highly related, so it might be a little DRYer and more future-proof to pull out a parent type, like:

```ts
interface BaseTaxData {
  readonly kind: DataType;
  readonly lastFourSsnDigits: string;
  readonly zipCode: string;
}

export interface AutofileData extends BaseTaxData {
  readonly kind: DataType.AUTOFILE;
  readonly paymentMethod?: PaymentMethod;
}

export interface StatusData extends BaseTaxData {
  readonly kind: DataType.STATUS;
  readonly applicationDateString: string;
  // ...
}

export type TaxReliefStatusData = AutofileData | StatusData;
```

webapp/components/TaxpayerInfoHeader.tsx
```diff
@@ -0,0 +1,44 @@
...
+        <div className={colSize}>
+          <p>
+            SSN/ITIN: <strong>***-**-{lastFourSsnDigits}</strong>
+          </p>
+        </div>
+        <div className={colSize}>
+          <p>
+            ZIP Code: <strong>{zipCode}</strong>
+          </p>
+        </div>{" "}
```
casewalker: dust/sand - Why is all the text here in `<p>` tags, but there is a loose space-char here floating between divs? Does it even do anything? I suspect it doesn't do anything and it should be removed, or if it does do something, it would be better to remove it and handle spacing with CSS. But please let me know if I am mistaken.

webapp/components/TaxpayerInfoHeader.tsx
casewalker, line 1: nit - This is at least a medium sized PR IMO. I think it would have been better to have separate PRs where one was focused on refactoring the `TaxpayerInfoHeader` alone (or maybe with the FAQs too) and the other handled the API work.

webapp/app/page.tsx
```diff
@@ -49,6 +69,25 @@ const hasTransactionWithStatus = (
   return false;
 };
 
+const hasPaymentSentTransaction = (record: StatusRecord) => {
+  return (
+    hasTransactionWithStatus(record.anchor, TransactionStatus.PAYMENT_SENT) ||
+    hasTransactionWithStatus(record.ptr, TransactionStatus.PAYMENT_SENT)
```
casewalker: dust/sand - I left this comment on a previous PR, IMO the for-loop in the separate function is unnecessary, this can be done in a clean one-liner:
```js
const hasPaymentSentTransaction = [...record.anchor, ...record.ptr].some((trx) => trx.status === TransactionStatus.PAYMENT_SENT);
```

webapp/app/page.tsx
```diff
@@ -30,11 +30,31 @@ export interface StatusRecord {
...
+}): Promise<AutofileResponse | null> => {
+  try {
+    const response = await fetch("/api/autofile", {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({ ssn: params.ssn, zip: params.zip }),
+    });
+
+    if (!response.ok) {
+      return null;
```
casewalker, line 49: dust/sand - Should this log anything anywhere for a non-200 response?

webapp/app/anchor-autofile/page.tsx
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  const { lastFourSsnDigits, zipCode, paymentMethod } = dataStore;
+  const paymentMethodString = paymentMethod === PaymentMethod.CHECK ? "Check" : "Direct Deposit";
+
+  return (
+    <main id="main-content">
+      <section className="usa-section">
+        <div className="grid-container">
+          <div style={{ textAlign: "right" }}>
```
casewalker, line 37: dust/sand - I don't think it is great mixing classes with raw `style` input, and I also think this should be able to be replaced with the class `text-right` if I am understanding [USWDS Utilities](https://designsystem.digital.gov/utilities/) correctly.

If that isn't supported, I think I would rather see a small CSS file with custom CSS classes rather than raw `style`.

webapp/app/anchor-autofile/AnchorAutofileDirectDepositFaqContent.tsx
```diff
@@ -0,0 +1,97 @@
+import { fireEventWhenFaqOpened, type FaqItem } from "@/components/FaqSection";
+
+export const AnchorAutofileDirectDepositFaqContent: FaqItem[] = [
+  {
+    title: "What if I did not receive a ANCHOR Benefit Confirmation Letter in the mail?",
```
casewalker: sand - `an ANCHOR`

webapp/app/anchor-autofile/AnchorAutofileCheckFaqContent.tsx
```diff
@@ -0,0 +1,97 @@
+import { fireEventWhenFaqOpened, type FaqItem } from "@/components/FaqSection";
+
+export const AnchorAutofileCheckFaqContent: FaqItem[] = [
+  {
+    title: "What if I did not receive a ANCHOR Benefit Confirmation Letter in the mail?",
```
casewalker: sand - `an ANCHOR`

webapp/app/anchor-autofile/AnchorAutofileDirectDepositFaqContent.tsx
casewalker, line 1: dust - It is pretty suspicious the amount of copy-paste between the two files AnchorAutofileDirectDepositFaqContent.tsx‎ and AnchorAutofileCheckFaqContent.tsx‎ (like, notice how I had to note the same `a ANCHOR...` typo twice?). James likes to sometimes go by the rule of three: once it is copy-pasted in 3 places, then it is likely time to DRY it out... but so much of these files are identical that it feels like it might be worth consolidating before that threshold.

Leaving this as a dust, it's up to you.

lambda-anchor-autofile-api/src/index.test.ts
```diff
@@ -35,12 +35,6 @@ describe("handler", () => {
       expect(result).toEqual({ error: "Request must include ssn (9 digits) and zip (5 digits)" });
     });
 
-    it("returns error for null event", async () => {
```
casewalker, line 38: question - Why remove this test?

webapp/app/api/autofile/route.test.ts
```diff
@@ -0,0 +1,76 @@
+import { describe, it, expect, vi } from "vitest";
```
casewalker, line 1: dust - out-of-scope, but the project should be configured so the vitest globals don't need to be explicitly imported in tests.

webapp/app/api/autofile/route.test.ts
```diff
@@ -0,0 +1,76 @@
...
+const createRequest = (body: unknown): NextRequest =>
+  new NextRequest("http://localhost/api/autofile", {
+    method: "POST",
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify(body),
+  });
+
+describe("POST /api/autofile", () => {
+  it("returns 400 when ssn is missing", async () => {
+    process.env.NEXT_PUBLIC_AUTOFILE_LAMBDA_API_URL = "https://lambda.example.com/invoke";
```
casewalker, line 26: dust/sand - Could this be pulled out to the global context and defined there once, rather than repeated in each individual test?

webapp/app/api/autofile/route.ts
```diff
@@ -0,0 +1,85 @@
...
+  });
+
+  const signer = new SignatureV4({
+    service: "lambda",
+    region: "us-east-1",
+    credentials: defaultProvider(),
+    sha256: Sha256,
+  });
+
+  return (await signer.sign(request)) as HttpRequest;
```
casewalker, line 38: question/sand - Why does this need to be forcibly cast `as HttpRequest`? What type is actually coming out of `signer.sign()`? IMO, the `as` typecast should be used incredibly sparingly in the functional code (but it is basically fine to toss around as you please in tests). It is almost a TypeScript hard-reset, like "_Don't trust your lying eyes, use this type instead_", and it can easily obscure type issues and bugs.

Not sure if this would be one of them, but there have been situations where the type is wrong and I use `satisfies` instead of `as`, it validates the shape matches the desired type rather than `as` which overrides the compiler.

webapp/app/api/autofile/route.ts
```diff
@@ -0,0 +1,85 @@
...
+  let body: StatusRequestBody;
+  try {
+    body = (await request.json()) as StatusRequestBody;
+  } catch {
+    return NextResponse.json({ error: "Invalid request body." }, { status: 400 });
+  }
+
+  if (!body.ssn || !body.zip) {
+    return NextResponse.json({ error: "Both ssn and zip are required." }, { status: 400 });
+  }
```
casewalker, line 64: dust - Out-of-scope for this PR, just an idea, but this could be handled with something like Zod schema parsing, rather than force type-casting the request JSON `as StatusRequestBody`, you could parse a defined shape, and you could have Zod provide the error-string here from the parse operation.

webapp/app/api/autofile/route.ts
```diff
@@ -0,0 +1,85 @@
...
+    const lambdaResponse = await fetch(lambdaUrl, {
+      method: signed.method,
+      headers: signed.headers,
+      body: signed.body,
+    });
+
+    const lambdaBody: unknown = await lambdaResponse.json();
+
+    return NextResponse.json(lambdaBody, { status: lambdaResponse.status });
+  } catch {
```
casewalker: pebble - This looks like it should probably have a defined error in the catch block, `catch (error)`, and then log that error somewhere useful.

casewalker (commented):
Neat stuff!

infra/lib/infra-stack.ts
jachan: You're right! I wasn't sure about this but tested in dev and consolidating this works.

webapp/components/TaxpayerInfoHeader.tsx
jachan: Unnecessary, removed! Thanks.

webapp/components/TaxpayerInfoHeader.tsx
jachan: Feedback taken, thank you for the note, will try to avoid in the future!

webapp/app/page.tsx
jachan: The separate helper function was originally being used to check if the Transaction had status ISSUE_FLAGGED, but looking at the code for that, it seems like it's no longer used. Agree that we can simplify.

webapp/app/anchor-autofile/page.tsx
jachan: We have a separate PR pending to refactor the logout button and fix the styling - I have made a note there to fix this.

https://github.com/newjersey/tax-relief-status-checker/pull/225

lambda-anchor-autofile-api/src/index.test.ts
jachan: I removed it as the previous test here was LLM-generated, and I didn't think it was relevant because we're constructing the API traffic ourselves. FWIW, a null event causes our API to fail during validation.

webapp/app/api/autofile/route.test.ts
jachan: You've called this out before! I investigated this, and the reason is because of conflicting imports between Cypress and Vitest. I can add a backlog ticket to fix this, but for now I'd like to unblock this PR as the feature is time-sensitive (useless to people after Sept15, and trying to reach as many people as possible).

casewalker: Yes, I tried to note that this is out-of-scope for the current PR, and I left it as a dust, I hope that clearly indicates that this comment should be seen as non-blocking.

webapp/app/api/autofile/route.ts
jachan: Thanks, fixed.

webapp/app/api/autofile/route.ts
jachan: The type that AWS gives us out of `signer.sign` is .... a `Promise<HttpRequest>`. Appreciate your note about the potential here to obscure type issues/bugs.

`satisfies` doesn't quite work here, but I will treat `HttpRequest` with added caution in the future. I think it is the right Type to use here as AWS SigV4 is returning it.

webapp/app/api/autofile/route.test.ts
jachan: I was proposing a path forward to resolve this and a reasoning for why I want to treat it with that priority level. Agree that Dust makes it clear that it is nonblocking.

This is the ticket: https://github.com/newjersey/tax-relief-status-checker/issues/241

webapp/app/api/autofile/route.ts
casewalker: OOH, my initial feelings here were right, but my first assessment missed the main answer.

The source of this issue seems to be that both SignatureV4 and smithy/protocol-http have separate `HttpRequest` classes, annoying that the names are clobbered. Claude suggests importing the _interface type_ from protocol-http separately and using that:

```ts
import { HttpRequest } from "@smithy/protocol-http";
import type { IHttpRequest } from "@smithy/protocol-http"; // <-- New line!
// ...

const signRequest = async (config: {
  // ...
}): Promise<IHttpRequest> => { // <-- Different type output
  // ...
  return await signer.sign(request); // <-- Types are now happy
};
```

Another issue this surfaced for me, it seems you are using https://www.npmjs.com/package/@aws-sdk/signature-v4 which is deprecated and hasn't updated in 3 years, you should migrate to `@smithy/signature-v4`.

Also the webapp package.json uses exact versions for all other dependencies rather than a carat version range, I would suggest aligning your dependency versioning strategy.

jachan: Thanks Case! I had noticed that the `@aws-sdk/signature-v4` package was deprecated, but was procrastinating on that update. No time like the present to fix this!

jachan: Fixed in https://github.com/newjersey/tax-relief-status-checker/pull/243. Thank you again for the suggestion here, @casewalker ! I apologize if I was dismissive earlier, I realize now that the true concern not `HttpRequest` but using `as` to override the type. Will be more careful when reviewing LLM generated code in the future, I've generally just glossed over things that are type-only. Thank you for the push to be better.

casewalker: Thank you for following up!! I do apologize again for not communicating this as clearly and kindly as I could have, thank you for listening to me anyway.

Yes, IMO `as` is quite insidious, it looks harmless but is actually dangerous and powerful, and it is commonly not respected for how much can be hiding under its type conversions.

PR #243, opened 2026-08-20
[SHOW] Update sigv4 library

jachan:
## Link to ticket

No ticket for this tech debt issue

## LLM Disclaimer

- An LLM generated the first pass of the first commit to upgrade `@aws-sdk/signature-v4` with `smithy/signature-v4`

## What was done?

- Replace `@aws-sdk/signature-v4` library (deprecated) with `smithy/signature-v4`, AWS' open source maintained library
- Allow install scripts from cypress, oracledb, and esbuild to resolve npm install warnings
- Update the `api/route` with the `smithy` types to avoid having to cast the HttpRequest type. s/o @casewalker for the assistance.

## Accessibility
N/A

## Steps to test
`npm run tests`
`npx cypress run`

N/A

## Notes

Any additional information about changes to approach or new patterns? Any future development
considerations or refactorings?

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,5 +39,11 @@
     "lambda",
     "lambda-anchor-autofile-api",
     "infra"
-  ]
+  ],
+  "allowScripts": {
+    "esbuild@0.28.1": true,
+    "esbuild@0.28.2": true,
+    "oracledb@7.0.1": true,
+    "cypress@15.20.1": true
+  }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -163,9 +163,9 @@
       "license": "Apache-2.0"
     },
     "node_modules/@aws-cdk/cloud-assembly-schema": {
-      "version": "54.17.0",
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -39,5 +39,11 @@
     "lambda",
     "lambda-anchor-autofile-api",
     "infra"
-  ]
+  ],
+  "allowScripts": {
+    "esbuild@0.28.1": true,
+    "esbuild@0.28.2": true,
```
casewalker, line 45: sand - I like the idea of having these `allowScripts` be tied to a specific version, that protects the repo against like a supply-chain attack. I do think you should probably only have one `esbuild` version in the project at a time (unless the dependencies are really ugly under the hood), so I would recommend just keeping one of these lines at a time.

jachan: Yes! The root cause was that `esbuild` just released a new version 12 days ago, and we weren't using the latest, this SHIP PR fixes that. https://github.com/newjersey/tax-relief-status-checker/pull/244. Renovate created an automated PR to fix this, but we would have still had the stale pinned version in the package.json file.

It's these tiny tech debt PRs that are so hard to find time for!

PR #285, opened 2026-09-02
[SHOW] - add stay NJ payments

### LINKED ISSUE #280: Add Stay Payments
**UPDATED 09/03 - Will no longer distinguish between "regular" and "update" payments for Stay NJ given we can't reliably tell the difference from data we have.**

**ALSO UPDATED 09/03 - Need to add feature flag so payments don't show until Taxation releases them. For now this should be "set" to 02/09/2027.

**User Story**
As an NJ resident eligible to receive a Stay NJ payment, I want to see my Stay NJ benefit payments in the Status Checker so that I know when to look out for my check or direct deposit and how much I should be expecting. I also want to know which payments are "regular" v. "update" so multiple checks don't feel random or unexplained. 

**Requirements**
If a Stay payment is the first within each of the following time periods, it should be considered a "regular" payment and use this copy: "Check sent on [DATE] or "Direct deposit made on [DATE].":
* February 1, 2027 - April 31, 2027
* May 1, 2027 - July 31, 2027
* August 1, 2027 - October 31, 2027
* November 1, 2027 - January 31, 2028

If a Stay payment is NOT the first within the time period, it should be considered an "update" payment and use this copy: "Your benefit amount was adjusted. Another check was sent on [DATE]." Note: update payments are always sent by check, so there is no direct deposit version of this content.

**Acceptance Criteria**
* Stay NJ payments show on the Payment Info Page.
* Payments are correctly categorized as "regular" or "update" and show the associated content.
* Note: still waiting on write access to TAXU elf_saver_inquiry so we can add records to test this feature.

natekep:
## Link to ticket

This commit resolves [280](https://github.com/newjersey/tax-relief-status-checker/issues/280#issue-5296081933)

## LLM Disclaimer

## What was done?

- A resident is now able to view their Stay NJ payments, however we are no longer showing a transaction as an "Update" payment, and instead they will all be considered "Regular". (see Notes)
- Changed name of `showEarliestTransaction` to `showTransactionRow` because it is no longer just showing the earliest transaction, and will be used to display each Stay NJ transaction. 

## Accessibility

## Steps to test

- Download code locally and run `NEXT_PUBLIC_ENABLE_STAY="true" npm run dev` and `npm run cypress`

- Before

Screenshot 2026-09-02 at 4 33 33 PM

- After

Screenshot 2026-09-02 at 4 32 16 PM

## Notes
@emtwelp and I had a meeting to discuss the date formatting for the checks in the TAX db. During our conversation the following scenario came up. In the case where a person receives a check near the end of the associated quarter range, and then that person gets an update payment with a date in the following quarter, that payment would be considered a regular payment instead of an update payment (Ex: Resident received Q1 check on 4/30, updated Q1 check on 5/2, and Q2 check on 5/5. Application would display the two Q1 payments as a regular payments and the Q2 payment as an updated payment to the second Q1 check). The current state of the DB does not have a way to differentiate between regular and update without using the date of the check. We decided that for Stay NJ payments, every check received will be treated as a regular payment and get displayed as such. Two considerations made going into that decision: 1. Not that common that someone will receive an updated Stay payment, and 2. This allows for subsequent application years to not have to worry about the date range for quarterly payments changing, and can be used as is for later development. We are also considering adding a column to the db to be able to differentiate the updated payments. 

Info on what is a "Regular" vs "Update" payment:

If a Stay payment is the first within each of the following time periods, it should be considered a "regular" payment and use this copy: "Check sent on [DATE] or "Direct deposit made on [DATE].":
- February 1, 2027 - April 30, 2027
- May 1, 2027 - July 31, 2027
- August 1, 2027 - October 31, 2027
- November 1, 2027 - January 31, 2028

If a Stay payment is NOT the first within the time period, it should be considered an "update" payment and use this copy: "Your benefit amount was adjusted. Another check was sent on [DATE]." Note: update payments are always sent by check, so there is no direct deposit version of this content.

### DIFF
--- a/webapp/app/page.tsx
+++ b/webapp/app/page.tsx
@@ -57,7 +57,7 @@ const checkAutofile = async (params: {
 };
 
 const determineRoute = (record: StatusRecord): string => {
-  const hasPaymentSentTransaction = [...record.ptr].some(
+  const hasPaymentSentTransaction = [...record.ptr, ...record.stay_nj].some(
     (transaction) => transaction.status === TransactionStatus.PAYMENT_SENT,
   );
 
--- a/webapp/app/page.tsx
+++ b/webapp/app/page.tsx
@@ -57,9 +57,17 @@ const checkAutofile = async (params: {
 };
 
 const determineRoute = (record: StatusRecord): string => {
-  const hasPaymentSentTransaction = [...record.ptr, ...record.stay_nj].some(
+  let hasPaymentSentTransaction = [...record.ptr].some(
     (transaction) => transaction.status === TransactionStatus.PAYMENT_SENT,
   );
+
+  if (process.env.NEXT_PUBLIC_ENABLE_STAY) {
+    hasPaymentSentTransaction =
+      hasPaymentSentTransaction ||
+      [...record.stay_nj].some(
+        (transaction) => transaction.status === TransactionStatus.PAYMENT_SENT,
+      );
+  }
 
   if (hasPaymentSentTransaction) {
     return "/payment-info";
--- a/webapp/app/payment-info/page.tsx
+++ b/webapp/app/payment-info/page.tsx
@@ -31,7 +31,7 @@ export const getEarliestTransaction = (transactions: Transaction[]) => {
   return earliestTransaction;
 };
 
-export const showEarliestTransaction = (transaction: Transaction, taxProgram: TaxProgram) => {
+export const showTransactionRow = (transaction: Transaction, taxProgram: TaxProgram) => {
   if (!transaction?.payment_details) return null;
   return (
     <tr>
@@ -73,7 +73,7 @@ export const showProgramTransactions = (transactions: Transaction[], taxProgram:
 
   return (
     <>
-      {showEarliestTransaction(earliest, taxProgram)}
+      {showTransactionRow(earliest, taxProgram)}
       {remainingTransactions.map((transaction) => showUpdatedTransaction(transaction, taxProgram))}
     </>
   );
@@ -95,7 +95,7 @@ const PaymentInfoPage = () => {
     return null;
   }
 
-  const { lastFourSsnDigits, zipCode, ptr } = dataStore;
+  const { lastFourSsnDigits, zipCode, ptr, stay_nj } = dataStore;
 
   return (
     <main id="main-content">
@@ -117,6 +117,12 @@ const PaymentInfoPage = () => {
               {(() => {
                 if (ptr.length === 0) return null;
                 return showProgramTransactions(ptr, TaxProgram.PTR);
+              })()}
+              {(() => {
+                if (stay_nj.length === 0) return null;
+                for (const transaction of stay_nj) {
+                  return showTransactionRow(transaction, TaxProgram.STAY_NJ);
+                }
               })()}
             </tbody>
           </Table>
--- a/webapp/app/payment-info/page.tsx
+++ b/webapp/app/payment-info/page.tsx
@@ -120,9 +120,9 @@ const PaymentInfoPage = () => {
               })()}
               {(() => {
                 if (stay_nj.length === 0) return null;
-                for (const transaction of stay_nj) {
-                  return showTransactionRow(transaction, TaxProgram.STAY_NJ);
-                }
+                return stay_nj.map((transaction) =>
+                  showTransactionRow(transaction, TaxProgram.STAY_NJ),
+                );
               })()}
             </tbody>
           </Table>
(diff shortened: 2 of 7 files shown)


webapp/app/payment-info/page.tsx
jachan, line 68: [PEBBLE] This code is confusing to a reader. The name of the function makes it seem as if all programs use this logic, but on the `payment-info` page, `stay_nj` has been updated to bypass this logic.

```
return stay_nj.map((transaction) => showTransactionRow(transaction, TaxProgram.STAY_NJ));
```

I would rather update `showProgramTransactions` to have logic that does different things based on the provided `TaxProgram` parameter.

webapp/app/payment-info/page.tsx
```diff
@@ -118,6 +118,12 @@ const PaymentInfoPage = () => {
                 if (ptr.length === 0) return null;
                 return showProgramTransactions(ptr, TaxProgram.PTR);
```
jachan, line 127: [DUST] Clarifying that the intent is to continue to show "updated transaction" copy for PTR transactions but NOT for StayNJ?

webapp/app/page.tsx
```diff
@@ -57,7 +57,7 @@ const checkAutofile = async (params: {
 };
 
 const determineRoute = (record: StatusRecord): string => {
-  const hasPaymentSentTransaction = [...record.ptr].some(
+  const hasPaymentSentTransaction = [...record.ptr, ...record.stay_nj].some(
```
jachan: [PEBBLE] Is there any concern around launch timing for this? For example, we were asked to not launch ANCHOR until Sept 15th.

jachan: Talked with Emily and we want to keep this behind a feature flag, since selection won't begin until 2027. Upgraded to PEBBLE

webapp/app/payment-info/page.tsx
jachan: Talked with Emily and confirmed that this is the case: PTR and ANCHOR will have updated transactions, StayNJ will not show updated transactions.

jachan (commented):
Need to add a feature flag for StayNJ and suggesting a refactor to function names

natekep:
Going to ask on wednesday, but in case someone wants to push before then, is the feature flag named correctly on [this line](https://github.com/newjersey/tax-relief-status-checker/pull/285/changes#diff-ee9f2e86d12835dc397a51abd76b072de191566eab162c987653e35d359023a9R64)

webapp/app/payment-info/page.tsx
```diff
@@ -118,6 +126,10 @@ const PaymentInfoPage = () => {
...
               })()}
+              {(() => {
```
jachan, line 129: [PEBBLE] If I'm reading this correctly, it's currently possible for someone to be shown STAY_NJ transactions even if the feature flag is disabled.

Consider: someone has PTR `payment_sent` transactions so they get routed to this page. They also have STAY_NJ transactions so those are displayed, even though the feature flag is false. 

To fix this, I believe this should check the feature flag on line 130 before displaying to users.

webapp/app/page.tsx
```diff
@@ -57,10 +57,18 @@ const checkAutofile = async (params: {
 };
 
 const determineRoute = (record: StatusRecord): string => {
-  const hasPaymentSentTransaction = [...record.ptr].some(
+  let hasPaymentSentTransaction = [...record.ptr].some(
     (transaction) => transaction.status === TransactionStatus.PAYMENT_SENT,
   );
 
+  if (process.env.NEXT_PUBLIC_ENABLE_STAY) {
```
jachan: [SAND] I find this pattern of assigning `hasPaymentSentTransaction` and then having this conditional block modify it to be pretty difficult to read.

Would rather use the feature flags to build an array of `recordsToCheck`, and then 

```
let recordsToCheck = [...record.ptr] // PTR is enabled by default
if (process.env.NEXT_PUBLIC_ENABLE_STAY) {
  recordsToCheck.append([...record.stay_nj])
}
let hasPaymentSentTransaction = recordsToCheck.some(
    (transaction) => transaction.status === TransactionStatus.PAYMENT_SENT,
  );

```

Off the top of my head, something like this seems easier to read and easier to extend in the future.

jachan (commented):
Left some comments here on feature flag implementation.

webapp/app/utils/determineRoute.ts
jachan, line 1: [PEBBLE] Can we add a test to `determineRoute.ts` to capture this new behavior?

PR #289, opened 2026-09-04
286 - change autofile api ordering

### LINKED ISSUE #286: Change routing to check elf_saver_inquiry before autofile data dump
**User Story**
As a resident who submitted their own ANC-1 application (or submitted a PAS-1 application before the ANCHOR autofile mailing list was compiled), I want the status checker to show me that an application is on file v. being told the state will file on my behalf so that I have peace of mind that my manual submission was successfully received and know I don't have to submit another application.

**Requirements**
* App should check elf_saver_inquiry for SSN & ZIP pairs before checking data dump and follow [this logic for pairs not found in elf_saver_inquiry](https://app.mural.co/t/officeofinnovation7132/m/officeofinnovation7132/1773178747648/ece9436aa3ca91e73f17217e3972eabfbd9f8a0e?wid=0-1788378146632).
* Pairs found in elf_saver_inquiry can follow existing logic.

**Acceptance**
* Residents with a record in BOTH elf_saver_inquiry and the autofile data dump should see App Received Page not ANCHOR Autofile Notice Page.
* Residents with a record ONLY  in the autofile data dump should see ANCHOR Autofile Notice Page. 
* Residents with ONLY a record in elf_saver_inquiry should stay the same.

jachan:
## Link to ticket

This commit resolves #286 

## LLM Disclaimer

- An LLM helped with generating ideas for refactoring `onSubmit` to be more legible
- I wrote the unit test cases, an LLM filled them in, and then I worked to remove redundant mocks and change testing approach (e.g. the determine route tests used to be done by filling the form, but now they test the function directly)

## What was done?
- Previously, we called the `autofile` API first, and used `status` API as a fallback if no record was found
- This PR changes the ordering. Now, `status` API will the be first call, and `autofile` will only be called if no record is found.
- Refactored `onSubmit` handler to be more readable
- added unit tests for landing page
- moved StatusRecord to shared types
- refactor to break out determineRoute into its own module

## Accessibility

## Steps to test
- Mock out both calls to return `record found` using browser tools, and verify that they are routed to the correct page rather than the `autofile` page.

Not a visual change

## Notes
- May be helpful to switch to a split diff in the GH review, at least for the updated `onSubmit` logic

### DIFF
--- a/webapp/app/page.test.tsx
+++ b/webapp/app/page.test.tsx
@@ -1,11 +1,12 @@
 import { describe, it, expect, vi, beforeEach } from "vitest";
 import { render, screen } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
-import LandingPage, { determineRoute } from "./page";
-import type { StatusRecord } from "./page";
+import LandingPage from "./page";
+import type { StatusRecord } from "@/components/types";
 import { TaxReliefDataProvider } from "@/components/TaxReliefDataProvider";
-import { PaymentMethod, TransactionStatus } from "@/components/types";
+import { PaymentMethod } from "@/components/types";
 import { logGAEvent } from "./utils/analytics";
+import { determineRoute } from "./utils/determineRoute";
 
 const mockPush = vi.fn();
 
@@ -15,6 +16,10 @@ vi.mock("next/navigation", () => ({
 
 vi.mock("./utils/analytics", () => ({
   logGAEvent: vi.fn(),
+}));
+
+vi.mock("./utils/determineRoute", () => ({
+  determineRoute: vi.fn(),
 }));
 
 const renderLandingPage = () =>
@@ -68,7 +73,7 @@ describe("onSubmit handler", () => {
       "/api/status",
       expect.objectContaining({ method: "POST" }),
     );
-    expect(mockPush).toHaveBeenCalledWith("/application-received");
+    expect(determineRoute).toHaveBeenCalledWith(record);
     expect(logGAEvent).toHaveBeenCalledWith("api_200_record_found");
   });
 
@@ -161,32 +166,5 @@ describe("onSubmit handler", () => {
       expect(mockPush).not.toHaveBeenCalled();
       expect(logGAEvent).toHaveBeenCalledWith("autofile_api_error");
     });
-  });
-});
-
-describe("determineRoute", () => {
-  it("routes a user to payment info when record has PTR payment sent", () => {
-    const record = buildStatusRecord({
-      ptr: [{ status: TransactionStatus.PAYMENT_SENT }],
-    });
-
-    expect(determineRoute(record)).toBe("/payment-info");
-  });
-
-  it("routes a user to more-information-needed when record has a flagged issue", () => {
-    const record = buildStatusRecord({
-      ptr: [{ status: TransactionStatus.ISSUE_FLAGGED, review_category: "SVR" }],
-      anchor: [{ status: TransactionStatus.ISSUE_FLAGGED, review_category: "MOD" }],
-    });
-
-    expect(determineRoute(record)).toBe("/more-information-needed");
-  });
-
-  it("routes a user to the application received page when no PTR payment sent and no issue is flagged", () => {
-    const record = buildStatusRecord({
-      ptr: [{ status: TransactionStatus.PROCESSING }],
-    });
-
-    expect(determineRoute(record)).toBe("/application-received");
   });
 });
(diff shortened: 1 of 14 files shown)


webapp/cypress/e2e/autofile.cy.ts
```diff
@@ -9,10 +9,26 @@ beforeEach(() => {
...
+  cy.intercept("POST", "/api/autofile", {
+    statusCode: 200,
+    fixture: "autofile_api_check.json",
+  });
+  cy.contains("button", `Check Status`).click();
+  cy.url().should("include", "/payment-info");
 });
 
 it("should route the user to the autofile page when autofile API returns true with CHECK", () => {
-  fillFields();
```
jachan, line 15: This call was redundant, as it was already in the `beforeEach` block

webapp/cypress/e2e/applicationReceived.cy.ts
```diff
@@ -35,10 +35,6 @@ beforeEach(() => {
     win.gtag = cy.stub().as("gtag");
   });
   cy.visit("/");
-  cy.intercept("POST", "/api/autofile", {
```
jachan, line 38: Previously, to properly test routing logic, we needed to assert that `autofile` API returned `planned: false`, as calling the status API was the fallback logic. 

Now, we've switched it so that calling autofile is the fallback logic, so this mock is no longer necessary. This applies to several of the Cypress tests.

webapp/app/page.tsx
```diff
@@ -33,30 +33,47 @@ interface AutofileResponse {
...
+  readonly ssn: string;
+  readonly zip: string;
+}): Promise<StatusResponse> => {
+  const response = await fetch("/api/status", {
+    method: "POST",
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify({ ssn: params.ssn, zip: params.zip }),
+  });
+
+  if (!response.ok) {
```
natekep, line 61: [DUST] - Just for my own understanding, does this get logged into GA4 only if there is an API code outside of 200-299? For instance, if records is empty (api_200 response), then the anchor autofile gets logged to ga4 as `autofile_${autofileResult.paymentMethod}` and not the `api_error` (assuming no error with the autofile)?

webapp/app/page.test.tsx
```diff
@@ -0,0 +1,192 @@
...
+    json: () => Promise.resolve(body),
+  }) as unknown as typeof fetch;
+
+beforeEach(() => {
+  vi.restoreAllMocks();
+  mockPush.mockReset();
+});
+
+describe("onSubmit handler", () => {
+  it("calls status API, if record found, sets data store and calls determineRoute", async () => {
```
natekep, line 65: [SAND] Up to you if you think this is valuable. I know we have unit tests for `determineRoute` later on, but should we also test the various routes a user could be pushed to within the onSubmit tests? (right now we only ensure it pushes to `/application-received`

webapp/cypress/e2e/autofile.cy.ts
```diff
@@ -9,10 +9,26 @@ beforeEach(() => {
   });
   cy.visit("/");
   fillFields();
+  cy.intercept("POST", "/api/status", {
+    statusCode: 200,
+    fixture: "v2_api_empty_records.json",
+  });
+});
+
+it("should route the user to the payment-info page when autofile API and status API return records", () => {
```
natekep, line 18: [SAND] - A little confused on this test. If the status api is called first, doesn't it only call the autofile api if there is no record found? In what situation would both return a record?

natekep (approved):
Looks really good! Much easier to read/understand the `onSubmit` handler now. Left some questions and a concern

webapp/cypress/e2e/autofile.cy.ts
jachan: You're right that only the status API will be called in our current flow.

I am mocking both to simulate a case where both APIs would return a record found. See ticket #286 acceptance criteria:

> Residents with a record in BOTH elf_saver_inquiry and the autofile data dump should see App Received Page not ANCHOR Autofile Notice Page.

webapp/app/page.tsx
jachan: Correct, we only want to log an `api_error` when a call to an API fails. We are not counting a response with an empty record as a failure.

webapp/app/page.test.tsx
jachan: Part of the reason I chose to abstract out the routing logic into its own function was so we could test it separately from the `onSubmit` logic. My thinking is that because we have written those tests, we can trust `determineRoute` to operate correctly and do not need to recreate those test cases here in the `onSubmit` section.

However, you bring up a good point. This test name is a little misleading, since we are not asserting `determineRoute` is called, but rather asserting which page the user gets routed to. This leads to the test being brittle, as if we change `determineRoute` logic in the future, this test might break. 

Let me fix that.

Issue #256, opened 2026-08-24
Implement ANC-1 Experience (ANCHOR-specific Processing and Payment Info Pages)


emtwelp:
**User story:** 
As a NJ resident who used the ANC-1 application to apply for 2025 property tax relief (whether I submitted one on my own or the state autofiled one on my behalf):
- If my application is processing, I want to be told that my application is being reviewed for ANCHOR only (which is not the current PAS-1 process), so that what I see matches my expectations about what my application was for, and I don't get confused.
- If a payment has been made for me, I don't be told about other programs on the payment page (which is the current PAS-1 design) so that I don't get confused about which programs I'm eligible for and should be expecting. 
- HOWEVER, if I accidentally used the ANC-1 when I shouldn't have, and the Division confirms this by paying me a benefit amount for Stay NJ or Senior Freeze, then I do want to want to be told about other programs.

**Requirements**
* API needs to be updated to account for new "form_code" indicator.
* New [ANC-1 Application Received](https://www.figma.com/design/fc59D8BV14x7EhW79ued5f/PAS-1-status-checker?node-id=1919-1292&t=nq0nkBtTBondDapR-4) and [ANC-1 Payment Info](https://www.figma.com/design/fc59D8BV14x7EhW79ued5f/PAS-1-status-checker?node-id=1919-1478&t=nq0nkBtTBondDapR-4) Pages created. 
* Routing should be updated so that users are shown: 1) either the ANC-1 or [PAS-1 Application Received Page](https://www.figma.com/design/fc59D8BV14x7EhW79ued5f/PAS-1-status-checker?node-id=1325-3212&t=nq0nkBtTBondDapR-4), and 2) either the ANC-1 or [PAS-1 Payment Info Page](https://www.figma.com/design/fc59D8BV14x7EhW79ued5f/PAS-1-status-checker?node-id=1510-3417&t=nq0nkBtTBondDapR-4) according to [this logic.](https://app.mural.co/t/officeofinnovation7132/m/officeofinnovation7132/1773178747648/ece9436aa3ca91e73f17217e3972eabfbd9f8a0e?wid=0-1788376303474) 

**Acceptance Criteria**
Check Processing:
* If SSN & ZIP are in elf_saver_inquiry, and resident used ANC-1 for their application (form_code = ANC-1), then they see ANCHOR App Received Page.
* If SSN & ZIP are in elf_saver_inquiry but form_code is NULL, then they see PAS-1 App Received Page.
[...]

PR #298, opened 2026-09-10
update API to return form_code

jachan:
## Link to ticket

This PR is the first step to resolving #256
The next PR will be an update to the webapp code to handle this new field.

## LLM Disclaimer

- An LLM was used to debug a type failure (root cause was that `enum` is executable code, and can't be in the `types.d.ts` file as esbuild won't read it). Resolution was manual, I moved the type into `index.ts`.

## What was done?

- We'd like to show users different content based on what their original filing form was
- Taxation has added a field to their DB that captures a taxpayer's original filing form
- This change updates the API to read this new field and include it in the API response

## Accessibility
N/A

## Steps to test

- `npm run test`
- This change has been deployed to our dev environment, connecting to the dev TAXU database, and we have verified that the behavior is correct by manually triggering the lambda with test cases.

None, this is an API change

## Notes
- Do reviewers feel like `form_code` in the response should be optional? Current API design expects the field to exist but allows it to be null when an unexpected value in the DB is encountered.

### DIFF
--- a/lambda/src/helpers.ts
+++ b/lambda/src/helpers.ts
@@ -13,5 +13,6 @@ export const buildMockRow = (overrides = {}) => ({
   CHECK_1_AMT: 1750,
   CHECK_1_NUM: "922775385",
   TRANS_1_TAX_CDE: 13,
+  FORM_CDE: "PAS1W",
   ...overrides,
 });
--- a/lambda/src/index.ts
+++ b/lambda/src/index.ts
@@ -29,9 +29,16 @@ interface ValidationResult {
   readonly zip?: string;
 }
 
+/** The form filed by the taxpayer for property tax relief */
+enum FormCode {
+  ANC1 = "ANC-1",
+  PAS1 = "PAS-1",
+}
+
 interface ResponseRecord {
   readonly return_year: string;
   readonly application_date: string;
+  readonly form_code: FormCode | null;
   readonly anchor: Transaction[];
   readonly ptr: Transaction[];
   readonly stay_nj: Transaction[];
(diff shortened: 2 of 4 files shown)


lambda/src/helpers.ts
josephine-njooi, line 1: Dust suggestion: renaming the file `testHelpers.ts` might help make it clearer without having to look into the file

lambda/src/index.ts
josephine-njooi, line 74: Dust: Not for this PR, but just a suggestion that if your validation starts getting even more complicated, you might want a consider a schema validation package like https://zod.dev/.
edit: It's also the suggested schema validation library in our [eng wiki](https://newjersey.github.io/innovation-engineering/tech-recommendations/backend/#api-type-safety) 🙂

It would allow you to declare what expected shape is, then let the package handle validating the whole shape at once. E.g., we [use this in Profile/SSO](https://github.com/newjersey/profile-nj/blob/75e1bb1f8d79256e66c0d7ef5e52347658f676e6/apps/profile/backend/src/utils/profile.ts#L41).

Zod can also do transforms ([e.g. this](https://github.com/newjersey/profile-nj/blob/75e1bb1f8d79256e66c0d7ef5e52347658f676e6/apps/profile/backend/src/utils/profile.ts#L41)), so the declaration could also call `mapFormCode` if you think the boundary makes sense.

lambda/src/index.ts
```diff
@@ -85,12 +92,33 @@ const mapRowToRecord = (row: InquiryRow): ResponseRecord => {
...
+      return FormCode.PAS1;
+    case "ANC1W":
+      return FormCode.ANC1;
+    case "ANC1D":
+      return FormCode.ANC1;
+    case "ANC1P":
+      return FormCode.ANC1;
+    default:
+      console.log(`Unknown FORM_CDE: ${dbFormCode}`);
+      return null;
```
josephine-njooi, line 118: > Do reviewers feel like form_code in the response should be optional? Current API design expects the field to exist but allows it to be null when an unexpected value in the DB is encountered.

I guess my question is, do you expect there to be other unexpected values in the DB?
- If no, that would seem like a validation error, similar to the SSN/ZIP validation errors
- If yes but they don't count, seems like null makes sense
- If you don't know, maybe it's worth a ticket to find out/make sure that unexpected values get alerted in production. And then to figure out if this is a sufficiently expected use case that you want to support this not existing in the frontend/whatever consumes this. Or, just error and report.

Is my 2c 😛

josephine-njooi (approved):

natekep:
> * Do reviewers feel like `form_code` in the response should be optional? Current API design expects the field to exist but allows it to be null when an unexpected value in the DB is encountered.

I agree with most of what Josephine said. Maybe this is me not having full context, but was the extra column added to every years record, or only 2025 and beyond? I feel like we ran into issues in the past of previous years having missing data, so allowing for null is preparing for when we show previous years imo.

natekep (approved):

jachan:
Running into GH issue where checks show as complete but are still blocking merge. Going to close and reopen this PR to try and fix it.

